PR #3, opened 2022-02-15
Init taipy-gui (#837)

dinhlongviolin1:


.github/workflows/doc.yml
```diff
@@ -3,7 +3,7 @@ name: Documentation Generation
 on:
   push:
     branches:
-      - develop
+      - main
```
FabienLelaquais, line 6: Why 'main', and do we need doc.yml here anyway?

taipy/__init__.py
```diff
@@ -3,14 +3,3 @@
 __author__ = """Avaiga"""
```
FabienLelaquais, line 3: I guess that 'taipy/__init__.py' will move to taipy-code or even taipy...

.github/workflows/test.yml
```diff
@@ -34,7 +34,7 @@ jobs:
         with:
           coverageFile: coverage.xml
           token: ${{ secrets.GITHUB_TOKEN }}
-          thresholdAll: 0.85
+          thresholdAll: 0.75
```
FabienLelaquais, line 37: I suspect that we should let the coverage as high as possible, even if this means a failure.
That's something we must improve...

FabienLelaquais (changes requested):
It looks like taipy-core code and tests are still hanging there, right?

.github/workflows/doc.yml
dinhlongviolin1: Just want to change it to main so that it wont get triggered right now once we push to develop branch since we will be changing the docs generation between multiple repos anyway

.github/workflows/test.yml
dinhlongviolin1: I will make it my prority to improve the tests this week after doing some PR for Florian Jacta. For now, I would prefer to keep it at 75% so that all future tests will pass. I will make a seperate PR for the tests and also edit the threshold back to 85% once I publish that tests pr

taipy/__init__.py
dinhlongviolin1: I am open to discussion on what to do with the __init__.py and __main__.py files

FabienLelaquais (approved):
Great great work!

FredLL-Avaiga (approved):
🥇

Issue #18, opened 2022-01-25
Rename the `date_selector` control to `date` if it can be turned to a static control


FabienLelaquais:
**Description**

The `date_selector` control could benefit from having an `editable` property, set to `True` by default.

If this happens, we will want to rename that control to make its name shorter and reflect better what it really does.
`date` is an option, but maybe `datetime` is better. Let's discuss.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #16, opened 2022-02-14
protect data


FredLL-Avaiga:
**Description**
For better robustness, it would be nice to protect Gui data by
- [ ] providing user data scope as a parameter to user function (instead of gui)
- [ ] user data scope should only allow access to the user python script variables

gui parameter will not be provided to user functions anymore:
- [ ] UI methods (block, send alert, navigate ...) can be called through the user created gui instance

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #15, opened 2022-02-13
Improve Encapsulation


FredLL-Avaiga:
**Description**
- [ ] Simplify holder evaluation
- [ ] add content and image encapsulation
- [ ] manage lov value selection

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #14, opened 2022-02-11
Allow early variable binding


FabienLelaquais:
**Description**

At the current time, variable binding is done when a page is first displayed. Only the variable used in that page would be bound, and therefore accessible from the Gui instance (or to be more precise, from the client data scope).

Let's suppose we have an application with two pages:
  - the app starts by showing page 1.
  - variable 'x' is used on page 2 only.
  - some functions triggered by buttons or controls of page 1 need to update 'x'
    Naturally, we want to refer to 'gui.x' and do our thing.
 Unfortunately, until page 2 is displayed, 'x' does not exist in the Gui object, and the application fails.
 
In the situation of some user code accessing a variable from the Gui object, but fails to find it (it is not bound), no matter if we want to read or write this variable, the Gui object should try to locate the not-yet-bound variable in the global scope (just at it does in the regular magic binding from the page renderers), and bind it if it can.
 
Only when the variable actually cannot be found should the application complain and fail.

Of course, we may want to cache the binding failures if there's a performance hit or, which I think is more relevant, if we don't want to generate the same error message over and over.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
a workaround would be to modify the global variable while the gui variable are not available

FabienLelaquais:
> a workaround would be to modify the global variable while the gui variable are not available

Then you as a developer would need to know whether or not 'page 2', in the example, had been displayed at one point... and use the appropriate variable accordingly. 'x', or 'gui.x'... Hmm does not work for me.

FredLL-Avaiga:
> > a workaround would be to modify the global variable while the gui variable are not available
> 
> Then you as a developer would need to know whether or not 'page 2', in the example, had been displayed at one point... and use the appropriate variable accordingly. 'x', or 'gui.x'... Hmm does not work for me.

I do agree

PR #32, opened 2022-02-15
Protect data & co.

FredLL-Avaiga:
- #16 protect data (State)
- #15 encapsulation (fix parameter.date support)
- #14 early variable binding
- #10 easy charts and tables ```<{[26,74,65]}|chart|type=pie|>```
- #18 date_selector -> date
- #4 vertical indicator ```<|MAE: {mae}|indicator|value={val}|min=20|max=-20|orientation=vertical|>```


gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -63,22 +70,37 @@ const DateSelector = (props: DateSelectorProps) => {
...
-        />
+    return (
+        <Box id={id} className={className} sx={boxSx}>
+            {editable ? (
+                withTime ? (
+                    <DateTimePicker
+                        value={value}
+                        onChange={handleChange}
+                        renderInput={renderInput}
+                        className={getSuffixedClassNames(className, "-picker")}
```
FabienLelaquais, line 81: The code is used three times. Better to factorize?
And... not too sure about the 'picker' part...

gui/src/components/Taipy/DateSelector.tsx
```diff
@@ -63,22 +70,37 @@ const DateSelector = (props: DateSelectorProps) => {
...
+                        disabled={!active}
+                    />
+                )
+            ) : (
+                <Field
+                    dataType="datetime.datetime"
+                    defaultValue={props.defaultDate}
+                    value={props.date}
+                    format={props.format}
+                    id={id + "-field"}
```
FabienLelaquais, line 99: Why not use 'id' at this point? The input field is not rendered...

tests/taipy/gui/control/test_date_selector.py
```diff
@@ -3,29 +3,29 @@
 from taipy.gui import Gui
 
 
-def test_date_selector_md_1(gui: Gui, helpers):
+def test_date_md_1(gui: Gui, helpers):
```
FabienLelaquais, line 6: You may even want to rename the file itself at this point...

taipy/gui/utils/_state.py
```diff
@@ -0,0 +1,36 @@
+import typing as t
```
FabienLelaquais, line 1: That really looks great. Eager to see that in action!
Should the Gui object know about those variables though?

taipy/gui/data/array_dict_data_accessor.py
```diff
@@ -0,0 +1,35 @@
...
+
+    __types = (dict, list, tuple)
+
+    @staticmethod
+    def get_supported_classes() -> t.List[str]:
+        return [t.__name__ for t in ArrayDictDataAccessor.__types]
+
+    def __get_dataframe(self, value: t.Any) -> pd.DataFrame:
+        if len(value):
+            if isinstance(value[0], (str, int, float, bool)):
```
FabienLelaquais, line 19: Shouldn't we test is all values are kind of the same type?
Or maybe the DataFrame constructor does it?

FabienLelaquais (changes requested):
Excellent, excellent job.
A few remarks that might need tiny changes...

gui/src/components/Taipy/DateSelector.tsx
FredLL-Avaiga: Wrong copy/paste, it should only be used 2 times with the same suffix

FredLL-Avaiga: And in different cases

gui/src/components/Taipy/DateSelector.tsx
FredLL-Avaiga: Id is used at the parent level

tests/taipy/gui/control/test_date_selector.py
FredLL-Avaiga: Will do

taipy/gui/utils/_state.py
FredLL-Avaiga: ?
It does

taipy/gui/data/array_dict_data_accessor.py
FredLL-Avaiga: I guess dataframe will throw an error if it's not happy 😁

gui/src/components/Taipy/DateSelector.tsx
FredLL-Avaiga: Do you have another proposal for picker?

tests/taipy/gui/data/test_arry_dict_data_accessor.py
```diff
@@ -0,0 +1,57 @@
+import pytest
```
dinhlongviolin1: File name typo

taipy/gui/gui.py
```diff
@@ -280,9 +284,11 @@ def __update_var(self, var_name: str, value: t.Any, propagate=True, holder: Taip
         elif holder:
             modified_vars.update(self._evaluate_holders(hash_expr))
         # TODO: what if _update_function changes 'var_name'... infinite loop?
-        if self.on_change:
+        if hasattr(self, "on_change") and isinstance(self.on_change, FunctionType):
```
dinhlongviolin1, line 287: Thanks

taipy/gui/gui.py
```diff
@@ -519,32 +525,31 @@ def __on_action(self, id: t.Optional[str], payload: t.Any) -> None:
         if action and hasattr(self, action):
             if self.__call_function_with_args(action_function=getattr(self, action), id=id, payload=payload):
                 return
-        if self.on_action:
+        if hasattr(self, "on_action") and isinstance(self.on_action, FunctionType):
```
dinhlongviolin1, line 528: Thanks again

dinhlongviolin1 (approved):
Great job! I just have 1 question. Since gui has been replaced by state, from what I understand, the users wont be able to access gui public method through state. Correct me if I am wrong

tests/taipy/gui/data/test_arry_dict_data_accessor.py
FredLL-Avaiga: Fixed

FredLL-Avaiga:
> Great job! I just have 1 question. Since gui has been replaced by state, from what I understand, the users wont be able to access gui public method through state. Correct me if I am wrong
That's true and it's "by design":
- the user variable are accessed via State
- the Gui behavior via Gui

dinhlongviolin1:
> > Great job! I just have 1 question. Since gui has been replaced by state, from what I understand, the users wont be able to access gui public method through state. Correct me if I am wrong
> > That's true and it's "by design":
> 
> * the user variable are accessed via State
> * the Gui behavior via Gui

From what I have seen in the code, users can't access gui directly from the parameters anymore right

FredLL-Avaiga:
> From what I have seen in the code, users can't access gui directly from the parameters anymore right

they should have an gui instance globally accessible ;-)

dinhlongviolin1:
> > From what I have seen in the code, users can't access gui directly from the parameters anymore right
> 
> they should have an gui instance globally accessible ;-)

That is exactly what I am expecting. If Fabien's ok with it, I am

FabienLelaquais (approved):
👍

Issue #6, opened 2022-02-14
BUG-Assign a full dictionary to a bound variable


FabienLelaquais:
**Description**

If the page content contains dotted-notation bindings (such `{my_dict.my_key}`), the Gui does not propagate the value to the user interface when you update the entire `gui.my_dict` object.

**How to reproduce**

```
from taipy.gui import Gui

d = { "v1": 0, "v2": 1}

def on_change(gui, var, val):
  d = { "v1": 2*val, "v2": val}
  gui.d = d
  
Gui(page="""
Value: <|{d.v1}|>

Slider: <|{d.v2}|slider|>
""").run()
```

**Expected behavior**
Moving the slider updates the `v2` property value.

In `on_change`, we expect that setting the new entire dictionary value will update `gui.v1` as well, but instead we get the error message: `on_change function invocation exception: 'dict' object has no attribute 'v1'`

PR #33, opened 2022-02-16
bug fix assign a full dictionary (#6)

dinhlongviolin1:
Solve bug #6

Instead of assignment (which is a real challenge in python to implement), i opted for the update function similar to python dict. below is the example code

```
from taipy.gui import Gui

d = { "v1": 0, "v2": 1}

def on_change(gui, var, val):
    if var == "d.v2":
        d = { "v1": 2*val, "v2": val}
        gui.d.update(d)
  
Gui(page="""
Value: <|{d.v1}|>

Slider: <|{d.v2}|slider|>
""").run()
```

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -34,16 +34,16 @@ from .taipyimage import TaipyImage
 from .types import WsType
 from .utils import (
     Singleton,
+    TaipyBase,
+    TaipyContent,
+    TaipyContentImage,
+    TaipyData,
+    TaipyLov,
+    TaipyLovValue,
     _get_non_existent_file_path,
     _is_in_notebook,
     _MapDictionary,
     attrsetter,
-    TaipyBase,
-    TaipyData,
-    TaipyLov,
-    TaipyLovValue,
-    TaipyContent,
-    TaipyContentImage,
     get_client_var_name,
 )
 from .utils._adapter import _Adapter
@@ -198,7 +198,7 @@ class Gui(object, metaclass=Singleton):
         def __getter(elt: Gui) -> t.Any:
             value = getattr(elt._get_data_scope(), name)
             if isinstance(value, _MapDictionary):
-                return _MapDictionary(value._dict, lambda s, v: elt.__update_var(name + "." + s, v))
+                return _MapDictionary(value._dict, lambda s, v: elt.__update_var(f"{name}.{s}", v))
             else:
                 return value
 
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -34,16 +34,16 @@ from .taipyimage import TaipyImage
 from .types import WsType
 from .utils import (
     Singleton,
-    TaipyBase,
-    TaipyContent,
-    TaipyContentImage,
-    TaipyData,
-    TaipyLov,
-    TaipyLovValue,
     _get_non_existent_file_path,
     _is_in_notebook,
     _MapDictionary,
     attrsetter,
+    TaipyBase,
+    TaipyData,
+    TaipyLov,
+    TaipyLovValue,
+    TaipyContent,
+    TaipyContentImage,
     get_client_var_name,
 )
 from .utils._adapter import _Adapter
--- a/taipy/gui/utils/MapDictionary.py
+++ b/taipy/gui/utils/MapDictionary.py
@@ -11,11 +11,8 @@ class _MapDictionary(object):
 
     local_vars = ("_dict", "_update_var")
 
-    def __init__(self, dict_import, app_update_var=None):
-        # Verify if dict_import is a dictionary
-        if not isinstance(dict_import, dict):
-            raise TypeError("should have a dict")
-        self._dict: dict = dict_import
+    def __init__(self, dict_import: dict, app_update_var=None):
+        self._dict = dict_import
         # Bind app update var function
         self._update_var = app_update_var
 
@@ -74,7 +71,7 @@ class _MapDictionary(object):
     def items(self):
         return self._dict.items()
 
-    def get(self, key: t.Any, value: None) -> t.Union[t.Any, None]:
+    def get(self, key: t.Any, value: t.Union[None, str] = None) -> t.Union[t.Any, None]:
         return self._dict.get(key, value)
 
     def clear(self) -> None:
--- a/taipy/gui/utils/MapDictionary.py
+++ b/taipy/gui/utils/MapDictionary.py
@@ -37,7 +37,8 @@ class _MapDictionary(object):
     def __setitem__(self, key, value):
         if self._update_var:
             self._update_var(key, value)
-        self._dict.__setitem__(key, value)
+        else:
+            self._dict.__setitem__(key, value)
 
     def __delitem__(self, key):
         self._dict.__delitem__(key)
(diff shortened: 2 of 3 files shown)


taipy/gui/gui.py
```diff
@@ -198,7 +198,7 @@ def __value_getter(self, name):
         def __getter(elt: Gui) -> t.Any:
             value = getattr(elt._get_data_scope(), name)
             if isinstance(value, _MapDictionary):
-                return _MapDictionary(value._dict, lambda s, v: elt.__update_var(name + "." + s, v))
+                return _MapDictionary(value._dict, lambda s, v: elt.__update_var(f"{name}.{s}", v))
```
FabienLelaquais, line 201: I recall that string concatenation is slightly faster than using f-strings... Any rationale here?

I just checked: this is true if (1) there are no more than two items to concatenate and (2) if items are strings.
We check both boxes.

taipy/gui/utils/MapDictionary.py
```diff
@@ -24,7 +29,7 @@ def __getitem__(self, key):
         value = self._dict.__getitem__(key)
         if isinstance(value, dict):
             if self._update_var:
-                return _MapDictionary(value, lambda s, v: self._update_var(key + "." + s, v))
+                return _MapDictionary(value, lambda s, v: self._update_var(f"{key}.{s}", v))
```
FabienLelaquais, line 29: Same comment on concat vs. f-strings

taipy/gui/utils/MapDictionary.py
```diff
@@ -7,12 +12,12 @@ class _MapDictionary(object):
     local_vars = ("_dict", "_update_var")
 
     def __init__(self, dict_import, app_update_var=None):
-        self._dict = dict_import
-        # Bind app update var function
-        self._update_var = app_update_var
         # Verify if dict_import is a dictionary
         if not isinstance(dict_import, dict):
             raise TypeError("should have a dict")
```
FabienLelaquais: Please improve this error message if possible - like a bit of context?

taipy/gui/utils/MapDictionary.py
```diff
@@ -69,23 +74,26 @@ def values(self):
...
+    def popitem(self) -> tuple:
         return self._dict.popitem()
 
-    def copy(self):
+    def copy(self) -> _MapDictionary:
         return _MapDictionary(self._dict.copy(), self._update_var)
 
-    def update(self):
-        return self._dict.update()
+    def update(self, d: dict) -> None:
```
FabienLelaquais, line 92: Looking good!

taipy/gui/utils/MapDictionary.py
```diff
@@ -1,3 +1,8 @@
+from __future__ import annotations
```
FabienLelaquais, line 1: What is this necessary?

FabienLelaquais (approved):
Yeah no way to avoid that.
And no better idea for the `update` wording.

I suspect there should be a unit test on this very specific use case.

taipy/gui/utils/MapDictionary.py
dinhlongviolin1: This is required if you want to reference your own class as a type

taipy/gui/utils/MapDictionary.py
dinhlongviolin1: Sure thing

FredLL-Avaiga: Is this test really necessary ?
We only instantiate _MapDictionnary when we have checked that value is a dict

taipy/gui/gui.py
dinhlongviolin1: Performance wise, although it is true that string concatnation is faster than f-string, it is only marginally faster. In this case, I don't mind if it is f string or concatnation. but f string does look more concise and readable to me

taipy/gui/utils/MapDictionary.py
FabienLelaquais: Whao. So that is news. Thanks.

taipy/gui/gui.py
FabienLelaquais: I won't fight on this one. You have a rationale, that's good enough for me.

taipy/gui/utils/MapDictionary.py
```diff
@@ -69,23 +74,26 @@ def values(self):
     def items(self):
         return self._dict.items()
 
-    def get(self):
-        return self._dict.get()
+    def get(self, key: t.Any, value: None) -> t.Union[t.Any, None]:
```
FredLL-Avaiga: What's the point of this None as a type ?

FredLL-Avaiga (changes requested):
I propose to remove the type test
and can you explain value: None ?

taipy/gui/utils/MapDictionary.py
```diff
@@ -69,23 +74,26 @@ def values(self):
...
-    def pop(self, key, default=None):
+    def pop(self, key, default=None) -> t.Any:
         return self._dict.pop(key, default)
 
-    def popitem(self):
+    def popitem(self) -> tuple:
         return self._dict.popitem()
 
-    def copy(self):
+    def copy(self) -> _MapDictionary:
```
FredLL-Avaiga, line 89: is this why the future import is needed ?

dinhlongviolin1: yes it is

taipy/gui/utils/MapDictionary.py
dinhlongviolin1: It is better for type hinting knowing that it can either be a None value, or any other value, although t.Any does contain None.

taipy/gui/utils/MapDictionary.py
dinhlongviolin1: yes it is not necessary in our case. we can just remove altogether and requires that dict_import is a dict

FredLL-Avaiga:
Should we prevent state.d = {"new_dict": "" ....} ?
I can do this in State

taipy/gui/utils/MapDictionary.py
```diff
@@ -69,23 +71,26 @@ def values(self):
     def items(self):
         return self._dict.items()
 
-    def get(self):
-        return self._dict.get()
+    def get(self, key: t.Any, value: t.Union[None, str] = None) -> t.Union[t.Any, None]:
```
FredLL-Avaiga: same as value: t.Optional[str] = None right ?
Should we call it default_value ?

FredLL-Avaiga (approved):
👍

taipy/gui/utils/MapDictionary.py
dinhlongviolin1: On second thought, the most correct should be t.Optional[t.Union[str, None]]. 

I do like default_value better than value.

I will make both changes and then merge

FredLL-Avaiga: I thought Optional brought None to the possibilities ?

dinhlongviolin1: Sorry I typed it out wrong due to the phone glitching. What i mean that optional is cleaner and shorter than the union way that I am doing. I will implement your suggesions :D

dinhlongviolin1:
@FredLL-Avaiga 

I have implemented your recommendation. Hope you are happy now :D

FredLL-Avaiga:
Always happy

Issue #7, opened 2022-02-13
Provide a PageRenderer checker


FabienLelaquais:
**What would that feature address**

When your Taipy application uses many or complex page structures, it can be difficult to spot problems in the page definitions.
Warning messages may stack in the console, sometimes without any location information (which page, which content, what line).

***Description of the ideal solution***

We could add some checking functionality that would provide the user with better information.

It could be a `check()` method in the `PageRenderer` class, so that would work for Partials as well.

This method could return the renderer page just as it would be sent to the client, or `None` if any problem was spotted.

***Caveats***

The line number in the original text fragment that created the PageRenderer is often lost. We may need to plug it in the generated tags for error processing purposes.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #40, opened 2022-02-17
Page Validator/Checker (#7)

dinhlongviolin1:


taipy/gui/server.py
```diff
@@ -143,7 +147,14 @@ def _render_page(self) -> t.Any:
         if page is None:
             return (jsonify({"error": "Page doesn't exist!"}), 400, {"Content-Type": "application/json; charset=utf-8"})
         # TODO: assign global scopes to current scope if the page has been rendered
-        page.render()
+        with warnings.catch_warnings(record=True) as w:
+            page.render()
+            if len(w) > 0:
+                print(message:= f"--- {len(w)} warnings were found for page {page.route} ---")
```
FabienLelaquais, line 151: You will want to use the Flush parameter set to True in order to timely generate the messages.
In the console, 'regular' errors appear with a specific color (red for errors). Do you think we can mimic that (using the old \033[ trick)?

taipy/gui/page.py
```diff
@@ -29,9 +30,13 @@ def __init__(self):
         self.route: t.Union[str, None] = None
         self.head: t.Union[str, None] = None
 
-    def render(self):
+    def render(self) -> None:
         if self.renderer is None:
-            raise RuntimeError(f"Can't render JSX for {self.route} due to missing renderer!")
+            warnings.warn(f"Can't render JSX for {self.route} due to missing renderer!")
```
FabienLelaquais, line 36: No need to expose JSX in this message. Please make it simpler for users that don't know or care. Usually both.

taipy/gui/renderers/__init__.py
```diff
@@ -39,15 +38,35 @@ def __process_content(self, content: str) -> None:
...
 
+class PageValidator(ABC):
+    """
+    This base class provide validation functionality to PageRenderer. This allow users to catch potential errors in runtime.
+    """
+
+    def __init__(self, page: PageRenderer) -> None:
+        self._page = page
+    
+    @abstractmethod
```
FabienLelaquais, line 80: Hum. True/False loos rough to me, although I'm not sure we need anything else (like the rendered page).
What I do know is that we will need to review our errors/warnings generation in the parsers in order to let things through (like unclosed block tags) and others not that much (like unbound variables).

FabienLelaquais (approved):
Good first steps!

taipy/gui/page.py
dinhlongviolin1: Thanks

FabienLelaquais: Sorry my message was not clear.
I would prefer f"Cannot render page {self.route}: no renderer found"

dinhlongviolin1: EVen less work for me now

dinhlongviolin1:
Replaced by another pr

Issue #23, opened 2022-01-03
BUG-Useless error message in Gui.__call_function_with_args()


FabienLelaquais:
**Description**

In situations where the user-defined action function raises an exception, the warning message is useless.

**Expected behavior**

Provide some information about what went wrong and where.

FredLL-Avaiga:
@FabienLelaquais What would you like ?
the message now is 
- `on_action '{action_function.__name__}' function invocation exception: {e}` if an exception is raised
- `Wrong signature for action '{action_function.__name__}'` if the function has more that 4 parameters
I guess that last case could be dealt with

FredLL-Avaiga:
current message is : 

> on_action 'do_something' function invocation exception: Local script has no attribute 'toto'

PR #42, opened 2022-02-17
rework gui.on_action & on_change and actions from controls (#23)

FredLL-Avaiga:
Try to run actions if possible by:
- getting defaults value from script or from Gui
- adapting the calls to argcount

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -252,12 +252,18 @@ class Gui(object, metaclass=Singleton):
         # TODO: what if _update_function changes 'var_name'... infinite loop?
         if hasattr(self, "on_change") and isinstance(self.on_change, FunctionType):
             try:
-                self.on_change(
-                    self.__state, var_name, value.get() if isinstance(
+                argcount = self.on_change.__code__.co_argcount
+                args = [None for _ in range(argcount)]
+                if argcount > 0:
+                    args[0] = self.__state
+                if argcount > 1:
+                    args[1] = var_name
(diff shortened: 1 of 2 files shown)


taipy/gui/gui.py
```diff
@@ -699,8 +698,18 @@ def bind_var_val(self, var_name: str, value: t.Any) -> bool:
         return False
 
     def __bind_local_func(self, name: str):
-        if isinstance(self.__locals_bind.get(name, None), FunctionType):
-            setattr(self, name, self.__locals_bind[name])
+        func = getattr(self, name, None)
+        if func is not None and not isinstance(func, FunctionType):
+            warnings.warn(
+                f"{self.__class__.__name__}.{name}: {func} should be a function; looking for {name} in the script.")
```
FabienLelaquais, line 704: Don't we have two warnings for the same problem at this point?

FabienLelaquais (approved):
Makes sense

taipy/gui/gui.py
FredLL-Avaiga: Here, we have 2 messages only if 
- Gui.on_change/on_action is defined but not a function
- script on_action/on_change is defined but not a function
And that's only at start

Then at runtime when calling an action

We could have a message at start and on call, if
- Gui.on_action is set and not a function
- script on_action doesn't exists
- no specific action is defined

I'll fix that

dinhlongviolin1 (approved):

Issue #30, opened 2021-09-01
Allow multiple instances of taipy.App


FabienLelaquais:
The current implementation of taipy.App has a bad flaw that needs fixing: the Markdown class instance that we use to parse the md template is for some reason not propagated to the extension code (like the `handleMatch()` methods of `InlineProcessor` subclasses).
Obviously, we want to access the App instance there, since this is where we can retrieve the _DataProbes_ that were created, and maybe more information (such as styling).

The workaround so far was to set a class variable (in `App.__init__()`) to the Markdown class.
This is pretty bad since therefore, you cannot have more that a single App in your application...

PR #44, opened 2022-02-19
multiple taipy instances (#30)

dinhlongviolin1:
Solve Issue #30

- [x] Refactoring
- [x] Parameterize Gui --> Gui can be initialized multiple times as separate instances
- [x] write test for multiple instance
- [x] fix github action tests not running via similar issue https://github.com/aws/aws-sam-cli/issues/3661
- [ ] fix some mypy typing issues from previous commits


taipy/gui/renderers/_markdown/__init__.py
```diff
@@ -7,20 +7,19 @@
...
-    def extendMarkdown(self, md):
-        md.registerExtension(self)
-        # md.preprocessors.add("taipy", Preprocessor(md), "_begin")
-        md.preprocessors.register(Preprocessor(md), "taipy", 210)
-        ControlPattern.extendMarkdown(md)
-        md.parser.blockprocessors.register(
-            StartBlockProcessor(md.parser), "taipy", 175
-        )  # process the block tags (layout and co)
-        md.treeprocessors.register(Postprocessor(md), "taipy", 200)
+    config = {"gui": ["", "Gui object for extension"]}
```
FredLL-Avaiga, line 13: Should this be made part of the instance ? ie self.config = ?

FredLL-Avaiga (approved):
Great job.
I only have one question :-)

taipy/gui/renderers/_markdown/__init__.py
FredLL-Avaiga: Or should I read more about Markdown extension ?

dinhlongviolin1: Oh this config must be defined by default so you can bind it later. This is a hack that I found that you can pass in parameter during extension initialization and it will be registered to self.gui if gui has been defined in the config. You can read the code from Extension class from Markdown since I dont believe this is documented anywhere

FredLL-Avaiga: but if you define it as static, won't it be a problem when  accessing from multiple instances ?

dinhlongviolin1: Config is static but during class initialization, the class initialize a local self.config based on the static one so the actual config will be self.config like you expected

FredLL-Avaiga: If it's not documented it might be worth an explanation as comment ;-)

dinhlongviolin1: Thanks! Will do!

taipy/gui/renderers/_markdown/__init__.py
```diff
@@ -7,20 +7,19 @@
...
 
+    def extendMarkdown(self, md):
+        from ...gui import Gui
 
-def makeTaipyExtension(*args: Any, **kwargs: Any) -> Extension:
-    return TaipyExtension(*args, **kwargs)
+        gui = self.config["gui"][0]
+        if not isinstance(gui, Gui):
+            raise RuntimeError("Gui instance is not bindded to Markdown Extension")
```
FabienLelaquais, line 20: "is not bindded to Markdown Extension" -> "is not bound to the Taipy Markdown extension"

FabienLelaquais (approved):
This looks perfect!
A tiny typo in a warning... that's all I've seen!

Issue #45, opened 2022-02-22
Improve expression evaluation and function binding +


FredLL-Avaiga:
**Description**
Allow for better expression evaluation with
- list comprehension
- module imports
- function usage

Eliminate most of function binding except for lambda who don't have a name

that would allow to write:
`<|{[math.cos(2 * math.pi * x / 100) * math.exp(-3 * x * decay / 50 / 360) for x in range(0, 720)]}|chart|>`

- change page_id to page for dialog and pane controls
- status should use tuple (type, message)

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #46, opened 2022-02-22
Many improvements (#45)

FredLL-Avaiga:
Allow for better expression evaluation with

- list comprehension
- module imports
- function usage
- Eliminate most of function binding except for lambda who don't have a name

that would allow to write:
`<|{[math.cos(2 * math.pi * x / 100) * math.exp(-3 * x * decay / 50 / 360) for x in range(0, 720)]}|chart|>`

- change page_id to page for dialog and pane controls
- status should use tuple (type, message)
- no default value for default property => part render = true by default


FabienLelaquais (approved):
All I seen looks just perfect to me.

dinhlongviolin1 (approved):

Issue #48, opened 2022-02-23
BUG-Cannot use lambda in action


FabienLelaquais:
**Description**

There are situation where you want to use a lambda as an action.

Say you just want to set a variable as the result of pressing a button.

The natural (?) way of doing this would be:
```
<|Set value|button|on_action={lambda: value = True}|>
```

Because Taipy GUI has this state object that represents the user variables, it could be, although cumbersome:
```
<|Set value|button|on_action={lambda s: s.value = True}|>
```

However, both syntaxes fail to be properly interpreted by Taipy.

**How to reproduce**
See how Taipy fails to interpret:

```
from taipy.gui import Gui

value = False

Gui(page="""
<|Set value|button|on_action={lambda: value = True}|>
""").run()
```

**Expected behavior**

The lambda expression is executed, and therefore the value is set.

Issue #22, opened 2021-12-20
Provide a NumPy data accessor


FabienLelaquais:
**What would that feature address**

Some customers may want to rely on NumPy only (and no Pandas).
At this point, Taipy can deal with Data Frames that are exposed to front-end controls such as tables or charts.
We must provide an equivalent for NumPy arrays as well.

***Description of the ideal solution***

Multi-dimensional NumPy arrays can be bound to Taipy controls just like we already do with Pandas' data frames.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #49, opened 2022-02-23
Numpy and actions

FredLL-Avaiga:
- #22 numpy data accessor
- access data column by index if no column names
- #48 lambda in actions
- all actions are now named on_*

### DIFF
--- a/taipy/gui/data/pandas_data_accessor.py
+++ b/taipy/gui/data/pandas_data_accessor.py
@@ -12,9 +12,12 @@ from ..gui import Gui
 
 
 class PandasDataAccessor(DataAccessor):
+
+    __types = (pd.DataFrame,)
+
     @staticmethod
     def get_supported_classes() -> t.List[str]:
-        return [pd.DataFrame.__name__]
+        return [t.__name__ for t in PandasDataAccessor.__types]
 
     @staticmethod
     def __style_function(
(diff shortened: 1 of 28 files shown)


taipy/gui/data/pandas_data_accessor.py
```diff
@@ -45,10 +48,10 @@ def __build_transferred_cols(
         self, gui: Gui, payload_cols: t.Any, data: pd.DataFrame, styles: t.Optional[t.Dict[str, str]] = None
     ) -> pd.DataFrame:
         if isinstance(payload_cols, list) and len(payload_cols):
-            col_types = data.dtypes[data.dtypes.index.isin(payload_cols)]
+            col_types = data.dtypes[data.dtypes.index.astype("string").isin(payload_cols)]
```
FabienLelaquais: The (numpy) doc says that we should use 'astype(str)'. I wonder if that is not a bit more efficient and less error-prone...

taipy/gui/renderers/builder.py
```diff
@@ -56,16 +56,17 @@ def __init__(
...
-            if not isinstance(properties_dict, _MapDict):
+            (prop_dict, prop_hash) = self.__parse_attribute_value(self.__attributes["properties"])
+            if prop_hash is None:
+                self.__gui.bind_var(prop_dict)
+                prop_dict = getscopeattr(self.__gui, prop_dict)
+            if not isinstance(prop_dict, _MapDict):
                 raise Exception(
-                    f"Can't find properties configuration dictionary {properties_dict_name}!"
+                    f"Can't find properties configuration dictionary {self.__attributes['properties'] if prop_hash is None else prop_hash}!"
                     f" Please review your GUI templates!"
```
FabienLelaquais: We don't want this line (and the '!' in the line above)

FabienLelaquais (approved):
Better, and better... well done!

dinhlongviolin1 (approved):
You are the best!!!! 👍 👍 💯 🥇

taipy/gui/data/pandas_data_accessor.py
FredLL-Avaiga: I'll try it

taipy/gui/renderers/builder.py
FredLL-Avaiga: What do you want then ?

FredLL-Avaiga: Changed this to a warning and rephrased

Issue #12, opened 2022-02-04
BUG-Block Markdown syntax does not work when not on an empty line


FabienLelaquais:
**Description**

If you need a block element in your markdown (part, layout...) then it must appear on a line on its own, which is unfortunate.

**How to reproduce**

```
from taipy.gui import Gui;
Gui(page = """
Some text <|
Some more text
|>
""").run()
```

Breaks.
Adding a carriage return before the opengin `<|` fragment does work though.

**Expected behavior**

The code above should be interpreted properly.

PR #62, opened 2022-02-25
block opening tag now works same line with other content (#12)

dinhlongviolin1:
Solve Issue #12 

This now works
```
from taipy.gui import Gui
Gui(page= """
a text <|
part content
|>
""").run()
```

### DIFF
--- a/taipy/gui/renderers/_markdown/preproc.py
+++ b/taipy/gui/renderers/_markdown/preproc.py
@@ -71,11 +71,12 @@ class Preprocessor(MdPreprocessor):
                     tag, properties = self._process_control(m.group(2), line_count)
                 if tag in MarkdownFactory._TAIPY_BLOCK_TAGS:
                     tag_queue.append((tag, line_count, m.group(1) or None))
-                    line = line[: m.start()] + "\n" + MarkdownFactory._TAIPY_START + tag + ".start"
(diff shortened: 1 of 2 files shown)


taipy/gui/renderers/_markdown/preproc.py
```diff
@@ -71,11 +71,12 @@ def run(self, lines: List[str]) -> List[str]:
                     tag, properties = self._process_control(m.group(2), line_count)
                 if tag in MarkdownFactory._TAIPY_BLOCK_TAGS:
                     tag_queue.append((tag, line_count, m.group(1) or None))
-                    line = line[: m.start()] + "\n" + MarkdownFactory._TAIPY_START + tag + ".start"
+                    new_line_delimeter = "\n" if line.startswith("<|") else "\n\n"
```
FabienLelaquais, line 74: Ooooh so that was the problem?

FabienLelaquais (approved):
Well well well... if that was it... no much to complain about.
Well spotted!!

taipy/gui/renderers/_markdown/preproc.py
dinhlongviolin1: Yes it is, 1 more /n and everything works. It is due to how block processor works

FredLL-Avaiga (approved):

Issue #17, opened 2022-02-14
Turn libmagic as optional


florian-vuillemot:
We put Libmagic as a mandatory package in our Pipfile and setup.py.
We want to turn it optional and let users install it with `pip install taipy[magic]` or something similar.

Interesting fact, if magic is not installed, the variable `_has_magic_module` is never set and an exception is raised on first read access.
```
if util.find_spec("magic"):
    import magic

    _has_magic_module = True
```

Note: On MacOs Libmagic should be installed with `brew` (https://pypi.org/project/python-magic/). Could be good to add this information as an error message if the user is on Mac.

**Acceptance criteria**
- [ ] Turn libmagic optional
- [ ] Update GitHub Action to run test with and without

PR #63, opened 2022-02-27
bug fix libmagic optional (#17)

dinhlongviolin1:
Solve Bug #17
- Make libmagic and rdp optional
- use util.find_spec in tests so tests work properly
- update setup.py
- tests with optional dependencies
- all tests suites now run in both ubuntu and windows with python 3.8, 3.9, 3.10

### DIFF
--- a/taipy/gui/data/utils.py
+++ b/taipy/gui/data/utils.py
@@ -1,9 +1,17 @@
 from __future__ import annotations
 
 import typing as t
+from importlib import util
 
 import numpy as np
-from rdp import rdp
+
+_has_rdp_module = False
+
+if util.find_spec("rdp"):
+    from rdp import rdp
+
+    _has_rdp_module = True
+
 
 if t.TYPE_CHECKING:
     import pandas as pd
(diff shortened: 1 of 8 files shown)


taipy/gui/data/utils.py
```diff
@@ -16,6 +24,8 @@ def df_data_filter(
     expected_row_count: t.Union[None, int],
     margin_of_error: float = 0.3,
 ) -> t.Tuple[pd.DataFrame, t.Union[None, str], str]:
+    if not _has_rdp_module:
+        return None
```
FabienLelaquais: I guess, if RDP is not there, that I would return the original dataframe...
Am I missing anything?

FabienLelaquais (approved):
Looking good except for this little question of mine...

taipy/gui/data/utils.py
dinhlongviolin1: Thanks for this, I returned the wrong thing :D

dinhlongviolin1: Updated

tests/taipy/gui/control/test_image.py
```diff
@@ -22,6 +24,9 @@ def test_image_file_md(gui: Gui, helpers):
             "<Image",
             'defaultContent="data:image/png;base64,',
         ]
+        if not util.find_spec("magic"):
+            expected_list = ["<Image", 'defaultContent="/taipy-content/taipyStatic0/TaiPyContent.', ".bin"]
+        print(expected_list)
```
FredLL-Avaiga: Remove this ?

FredLL-Avaiga (approved):

tests/taipy/gui/control/test_image.py
dinhlongviolin1: thanks

Issue #56, opened 2022-02-24
Tree: Manage expanded state of nodes 


FlorianJacta:
**Description**
Add a n expanded dynamic property to the tree control
- if False, allows only one node to be expanded at any given level (ie close an expanded node if the suer expand another one)
- if a list of ids, expand nodes with id in the list

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #66, opened 2022-02-28
tree expand nodes (#56)

FredLL-Avaiga:
if expanded is boolean value False: only one node can be expanded at one given level
else it should be an array of the ids that need to be expanded (dynamic)


FabienLelaquais (approved):
All this looks just perfect.

dinhlongviolin1 (approved):
Nicely done!

Issue #64, opened 2022-02-28
Separate download functionality and download control


FabienLelaquais:
**Description**

Allow users to trigger a download when they need it.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #67, opened 2022-03-01
Gui.download (#64)

FredLL-Avaiga:
```<|Download file|button|on_action={lambda s: gui.download("/downloads/fred.png")}|>```

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -763,7 +763,7 @@ class Gui:
         """Sends a notification to the user interface.
 
         Args:
-            type (string): The notification type. This can be one of `"success"`, `"info"`, `"warning"` or `"error"`.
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: taipy/gui/gui.py [hunk 1]
```diff
@@ -745,11 +766,11 @@ def show_notification(
             type (string): The notification type. This can be one of `"success"`, `"info"`, `"warning"` or `"error"`.
```
FabienLelaquais: (optional(string)), right?

FabienLelaquais (approved):
This looks really, really good!
Still my open question about whether this API (as well as some others) should be on the State... but that's another topic.

taipy/gui/gui.py
FredLL-Avaiga: yes indeed

dinhlongviolin1 (approved):
Thanks for the included test :D

Issue #8, opened 2022-02-10
Add an element tag identifier in Markdown parser


FabienLelaquais:
**Description**

When your Markdown input becomes large, with lots of parts and layout and whatnot, you tend to suffer when it comes to making sure you're breaking the document structure.

The Markdown parser helps indicate where blocks are not closed or where block closing fragments are misplaced... but you don't really have any support on opening and closing fragments match.

We could extend the visual element Markdown syntax by allowing an identifier to be placed between the '<' and the '|', expecting the same identifier in the related closing fragment that would become '|''[opening_id]''>'.

So this syntax would be allowed:
```
<first|
some content
<second|
some more
|second>
|first>
```

If the '|second>' fragment is omitted, the parser will be able to mention it precisely.
If '|first>' and '|second>' are swapped, the parser will indicate it as well (although theoretically harmless).

This would be optional, and the parser would then be able to generate relevant messages to the programmer in case she gets lost in the document structure.
This really makes sense only for block constructs, but it is harmless to make it possible to use for controls as well.
Only a subset of identifier would be allowed, to simplify the parsing. The character set [0-9a-zA-Z_.] seems sufficient.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #68, opened 2022-03-01
Fix build file

florian-vuillemot:
You can now install taipy-gui with:
```
pip install taipy-gui
```
Then run in your python interpreter:
```
import taipy.gui as gui
```

We can't have something like:
```
import taipy as gui
```
due to taipy package.

https://github.com/Avaiga/taipy/issues/8

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,7 @@ setup(
     include_package_data=True,
     keywords="taipy-gui",
     name="taipy-gui",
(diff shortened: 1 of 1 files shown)


FabienLelaquais (approved):
That's a strange limitation you indicate here... but well...
Honestly, I can live with it.
Good job.

dinhlongviolin1 (approved):
Thanks Florian V

FredLL-Avaiga (approved):
🥇

Issue #65, opened 2022-02-28
BUG-Find a way to workaround the Plotly redraw problem


FabienLelaquais:
**Description**

Redrawing Plotly charts may result in erroneous output, where the chart area is mixed with the other components.

FredLL-Avaiga:
@FabienLelaquais Do we have some reproduction code ?

Issue #20, opened 2022-01-17
Add hover text on controls


FabienLelaquais:
**Description**

Hover information provides tips on what a control does.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

AAstridVH:
I guess you could use a standard "tooltip" item.

Try, if possible, to respect [these criteria](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
)

The Material UI React [tooltip with HTML](https://mui.com/components/tooltips/#customization) may work great.

FredLL-Avaiga:
@FabienLelaquais How would you call that property (that would be available on **all** components) ?

FabienLelaquais:
What about hover_text?
I don't fancy _label...

FredLL-Avaiga:
> What about hover_text?
> I don't fancy _label...

hover_text it will be then

PR #69, opened 2022-03-02
hover text (#20)

FredLL-Avaiga:
- #20 hover text on all components
- support date & time as well as datetime
- allow lov as a list of date or datetime ...
- rename tp_varname as updateVarName
- rename tp_updatevars as updateVars
- #65 option to change responsive to False in plot_config property


gui/doc/shared.csv
```diff
@@ -3,3 +3,4 @@ id,str,"""""",The identifier that will be assigned to the rendered HTML componen
 properties,"dict[str, any]",None,Bound to a dictionary that contains additonnal properties for this element.
 class_name,str,None,"CSS class name that will be associated with the generated HTML Element.
 This class names will be added to the default <code>taipy-&lt;element_type&gt;</code>."
+hover_text,dynamic(str),None,"Hover information provides tips on what a control does."
```
FabienLelaquais, line 6: Actually... the text could be anything...
What about:
"Information that is displayed when the user hovers over this element"
?

dinhlongviolin1 (approved):
Looks good to me

taipy/gui/utils/date.py
```diff
@@ -1,19 +1,20 @@
...
-def date_to_ISO(date: datetime) -> str:
-    # return date.isoformat() + 'Z', if possible
-    try:
-        return date.astimezone(utc).isoformat()
-    except Exception:
-        # astimezone() fails on Windows for pre-epoch times
-        # See https://bugs.python.org/issue36759
-        warnings.warn("There is some problems with date parsing to ISO!")
-        return date.isoformat()
+def date_to_ISO(date_val: t.Union[datetime, date, time]) -> str:
```
FabienLelaquais, line 8: Shouldn't this be hidden from users?

FabienLelaquais (approved):
Great job.
As for the Chart thing... if it works, it works...

taipy/gui/utils/date.py
FredLL-Avaiga: it will in the next PR

Issue #51, opened 2022-02-23
Pytest for playwright


dinhlongviolin1:
* Taipy version:
* Python version:
* Operating System:

### Description

We already have testing for backend (Python) and frontend (React) individually. It would be great to also be able to test our application end-to-end. Playwright has been chosen by me (Long) over puppeteer for this purpose. The goal of this issue is to implement a structure for writing end-to-end testing using playwright on pytest.

PR #71, opened 2022-03-03
e2e testing (#51)

dinhlongviolin1:
Solve Issue #51

### DIFF
--- a/.github/workflows/setuptools.yml
+++ b/.github/workflows/setuptools.yml
@@ -53,7 +53,7 @@ jobs:
           rm -r taipy  # Removal of Taipy code to ensure the use of the installed package.
           # Run test with optionals packages installed
           pip install pytest pytest-mock
-          pytest -m "not teste2e"> pytest_res
+          pytest -m "not teste2e" > pytest_res
           cat pytest_res
           echo "Ensure that no tests are skipped"
           tail -n1 pytest_res | grep --invert-match "skipped"
--- a/.github/workflows/setuptools.yml
+++ b/.github/workflows/setuptools.yml
@@ -11,7 +11,6 @@ jobs:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [16.x]
         python-versions: ['3.8', '3.9', '3.10']
         os: [ubuntu-18.04, windows-latest]
 
@@ -22,31 +21,18 @@ jobs:
       - uses: actions/setup-python@v2
         with:
           python-version: ${{ matrix.python-versions }}
-
-      - name: npm build and test ${{ matrix.node-version }} on ${{ matrix.os }}
-        uses: actions/setup-node@v2
-        with:
-          node-version: ${{ matrix.node-version }}
-      - name: Install dependencies
-        working-directory: ./gui
-        run: npm ci
-      - name: Run npm build
-        run: npm run build --if-present
-        working-directory: ./gui
 
       - name: Install Taipy without dependencies
         run: |
           pip install .
           rm -r taipy  # Removal of Taipy code to ensure the use of the installed package
-          pip install pytest pytest-mock pytest-playwright
-          playwright install
-          pytest
+          pip install pytest pytest-mock
+          pytest -m "not teste2e"
 
   optionals-packages:
     timeout-minutes: 10
     strategy:
       matrix:
-        node-version: [16.x]
         python-versions: ['3.8', '3.9', '3.10']
         os: [ubuntu-18.04, windows-latest]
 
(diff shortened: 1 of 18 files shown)


tests/taipy/gui/e2e/test_redirect.py
```diff
@@ -0,0 +1,24 @@
...
+
+@pytest.mark.teste2e
+def test_redirect(page: "Page", gui: Gui, helpers):
+    page_md = """
+<|Redirect Successfully|id=text1|>
+"""
+    gui.add_page(name="test", page=page_md)
+    gui.run(run_in_thread=True, single_client=True)
+    while not helpers.port_check():
+        time.sleep(0.5)
```
FabienLelaquais, line 20: No better way here?

tests/taipy/gui/e2e/test_redirect.py
```diff
@@ -0,0 +1,24 @@
...
+    page.goto("/")
```
FabienLelaquais, line 21: I don't quite understand the magic of using 'page' here. Please explain that to me. Thx

.github/workflows/setuptools.yml
```diff
@@ -26,8 +26,9 @@ jobs:
         run: |
           pip install .
           rm -r taipy  # Removal of Taipy code to ensure the use of the installed package
+          pip install markupsafe==2.0.1
           pip install pytest pytest-mock
-          pytest
+          pytest -m "not teste2e"
```
FabienLelaquais, line 31: Is this temporary?

.github/workflows/test.yml
```diff
@@ -36,6 +36,40 @@ jobs:
...
+        run: npm ci
+      - name: Run npm build
+        run: npm run build --if-present
+        working-directory: ./gui
+
+      - name: Run tests
+        run: |
+          python -m pip install --upgrade pip
+          pip install pipenv
+          pipenv install markupsafe==2.0.1
```
FabienLelaquais, line 68: It looks like this is protected in Pipfile... is it not?

FabienLelaquais (approved):
This look really powerful! Well done

.github/workflows/setuptools.yml
dinhlongviolin1: No this is not temporary. This is to specify that you only want to run test that is not for the end-to-end testing

tests/taipy/gui/e2e/test_redirect.py
dinhlongviolin1: I tried and this is the best I was able to come up with

.github/workflows/test.yml
dinhlongviolin1: Weird thing, despite me putting in pipfile, there is still a problem of not being able to override the original version, which is higher and incompatible. so I have to install it manually.

tests/taipy/gui/e2e/test_redirect.py
dinhlongviolin1: the page is provided by pytest_playwright, it is created automatically for you for each test.

.github/workflows/test.yml
```diff
@@ -36,6 +36,40 @@ jobs:
           token: ${{ secrets.GITHUB_TOKEN }}
           thresholdAll: 0.75
 
+  e2e:
```
FredLL-Avaiga: I suppose it stands for end to end ? Maybe change the name to end-to-end ?

pytest.ini
```diff
@@ -0,0 +1,6 @@
+[pytest]
+filterwarnings =
+    ; ignore pytest warnings when gui thread is stopped during playwright testing
+    ignore::pytest.PytestUnhandledThreadExceptionWarning
```
FredLL-Avaiga, line 4: Would it be possible to handle that Exception ?

tests/taipy/gui/e2e/test_redirect.py
FredLL-Avaiga: Maybe make the sleep shorter ? (timeout to connect is 2 secs)

FredLL-Avaiga (approved):

tests/taipy/gui/e2e/test_redirect.py
dinhlongviolin1: i will change timeout to connect is 1 and leep to 0.2

.github/workflows/test.yml
dinhlongviolin1: will do!

pytest.ini
dinhlongviolin1: I tried to handle it, but since it is in a seperate thread, it is not caught by gui stop method. Please let me know if you have any luck with handling that exception

PR #75, opened 2022-03-03
Webapp generation during installation

florian-vuillemot:
https://github.com/Avaiga/taipy/issues/8


dinhlongviolin1:
This will run on the user's machine during installation right

florian-vuillemot:
If the user installs Taipy from Github. We will send the package already built on Pypi. We didn't validate this point yet.

Issue #81, opened 2022-03-07
BUG-the selected value of a selector returns a '_TaipyLovValue' and not the true value


FlorianJacta:
**Description**
When the selected value of a selector returns a '_TaipyLovValue' and not directly the value. However, 'var_value' in the 'on_change' seems to be the real value of the selection.

**How to reproduce**
```
from taipy.gui import Gui

selector = ['test 1', 'test 2']
selected = selector[0]

md = """
<|{selected}|selector|lov={selector}|>
"""

def on_change(state, var_name, var_value):
    if var_name == 'selected':
        print(var_value)
        print(state.selected)
        
Gui(page=md).run()
```

**Expected behavior**
When we call state.selected, we should get 'test_1' or 'test_2' depending on the selection and not the '_TaipyLovValue'.

**Runtime environment**
Please specify relevant indications.
 - OS: Windown 10, conda environment
 - Browser: Firefox

Issue #79, opened 2022-03-04
BUG-Chart Select Property generates 2 on_change calls


vhfgosselin:
**Description**
When selecting 1 point in a chart using the selected property, the on_change function is called twice instead of once.
The behavior is not seen when using a box select on the chart, it just occurs when a single selection of a point
is performed.

**How to reproduce**
A code fragment
```
from taipy.gui import Gui, Markdown
import pandas as pd
l=[]
my_df = pd.DataFrame(data={"date":[1,2,3,4,5], "Qty":[10,20,10,20,10]})
page1 = """
# Bug Screen
<|{my_df}|chart|x=date|y=Qty|height=40%|width=50%|selected={l}|>
"""

def on_change(gui, var, value):
    if var=="l":
        print(f"value of f: {value}")

list_pages = {
    'p1':Markdown(page1)
}
gui = Gui(__name__, Markdown(page1))
gui.on_change = on_change
gui.run()
```

**Expected behavior**
If I select 1 point once, only 1 print should appear.

**Screenshots**

Issue #78, opened 2022-03-04
Chart: allow multi-dataframe support (different length)


FredLL-Avaiga:
**Description**
Chart: allow multi-dataframe support (different length)

That would allow for vertical timeline on charts for example

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #82, opened 2022-03-07
Charts and lov

FredLL-Avaiga:
- #78 multi-data
- #79 multiple on_change event
- #81 don't show inner types in state


FabienLelaquais (approved):
That looks really awesome from the user's standpoint! Well done.

dinhlongviolin1 (approved):
Everything looks really good!

Issue #92, opened 2022-03-10
BUG-A change in dictionary element doesn't call the on_change function


FlorianJacta:
**Description**
The problem is that when changing an element of a dictionary, the on_change function is not called. The element can be a single value or a list.

**How to reproduce**
```
from taipy.gui import Gui

new_list = ['YES 1','YES 2',"YES 3"]

a = 5

d = {
    "selected_item":'item_1',
    "item_selector":['item_1','item_2','item_3'],
    'value':1
}

def on_change(state,var_name,var_value):
    print(var_name,var_value)
    if var_name == 'd.selected_item':
        state.d['item_selector'] = new_list 
        print(state.d['item_selector'])
        state.d['value'] = 15
        state.a = 15


Gui(page="<|{d.selected_item}|selector|lov={d.item_selector}|>").run()
```

**Expected behavior**
When d['value'] and d['item_selector'] are changing when clicking on the selector, they should call the on_change, and we should see the print() done by their calls. This problem can also be seen in the selector that is not directly changing in the client.

PR #93, opened 2022-03-10
#92 dict update

FredLL-Avaiga:
- Fix dict update
- improve state.assign to support dotted notation


FabienLelaquais (approved):
Now that you explained...

dinhlongviolin1 (approved):
Approve in the blink of an eye

PR #99, opened 2022-03-14
Augment Reference Manual

FabienLelaquais:


taipy/gui/gui_actions.py
```diff
@@ -6,9 +6,9 @@
 
 
 def download(state: State, content: t.Any, name: t.Optional[str] = "", on_action: t.Optional[str] = ""):
-    """Donwloads content to the client.
+    """Donwload content to the client.
```
FredLL-Avaiga, line 9: Donwload => Download

taipy/gui/page.py
```diff
@@ -0,0 +1,60 @@
+from abc import ABC, abstractmethod
+import typing as t
+from .utils import _varname_from_content
+from os import path
+
+
+class Page(ABC):
+    """
+    The base class that transform template text to actual pages that can be
```
FredLL-Avaiga: transforms

taipy/gui/page.py
```diff
@@ -0,0 +1,60 @@
...
+    displayed on a Web browser.
+
+    When a page is requested to be displayed, it is transformed into HTML
+    code that can be sent to the client. All control placeholders are
+    replaced by the appropriate graphical component so you can display
+    your application variables, and potentially interact with them.
```
FredLL-Avaiga: remove potentially ?

taipy/gui/partial.py
```diff
@@ -6,7 +6,18 @@
 
 
 class Partial(_Page):
```
FredLL-Avaiga, line 8: is it _Page or Page ?
If we still have a _Page and it is relevant, I think we should rename it to something not so close to Page ...

taipy/gui/gui.py
```diff
@@ -54,14 +55,20 @@
...
-            triggers an action, as the result of an interaction with the end-user.
+        on_action (Callable): The function that is called when a control
+            triggers an action, as the result of an interaction with the end-user.<br/>
+            It defaults to the _on_action_ global function defined in the Python
+            application. If there is no such function, actions will not trigger anything.
         on_change (Callable): The function that is called when a control
-            modifies the variable it is bound to, as the result of an interaction with the end-user.
+            modifies variables it is bound to, as the result of an interaction with the
+            end-user.<br/>
+            It defaults to the _on_action_ global function defined in the Python
```
FredLL-Avaiga: _on_action_ => _on_change_

taipy/gui/gui.py
```diff
@@ -80,33 +87,37 @@ class Gui:
...
+
+        Arguments:
+            page: An optional `Page^` instance that is used when there is a single page
+                in this interface, referenced as the _root_ page (located at `/`).<br/>
+                If _page_ is a raw string and if it holds a path to a readable file then
+                a `Markdown^` page is built from the content of that file.<br/>
+                If _page_ is a string that does not indicate a path to readable file then
+                a `Markdown^` page is built from that string.<br/>
+                Note that if _pages_ is provided, those pages are added as well.
+            pages: TODO
```
FredLL-Avaiga: link to the add_pages() method ?

FredLL-Avaiga (approved):
a few comments

dinhlongviolin1 (approved):
Everything looks great

taipy/gui/gui.py
FabienLelaquais: Will do

taipy/gui/page.py
FabienLelaquais: Well as was thinking... |text| or |part| won't let you interact. But these are exceptions.
You are right.

taipy/gui/page.py
FabienLelaquais: Good catch

taipy/gui/gui_actions.py
FabienLelaquais: Thanks!

taipy/gui/partial.py
FabienLelaquais: So that's the Gui-independent part of the rendering process...
It sounds like a page to me (actually more than what 'Page' does)... and I have no better idea.
Do you?

taipy/gui/gui.py
FabienLelaquais: Greeeat catch!

Issue #100, opened 2022-03-14
Allow url_prefix for Gui object


dinhlongviolin1:
**Description**
Currently, all of the flask routes on Gui is hosted on the root routes. We might want to add an option so the user can host it on a secondary route so they can use the root route for other purposes.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #106, opened 2022-03-15
url prefix (#100)

dinhlongviolin1:
Solve Issue #100

Please have a careful look at this PR. It touches a lot of important files :D

- Flask version is now >=2.0
- Links, ws, auto-redirect on root page are all working with url_prefix
- Added a few tests (e2e) with url prefix


gui/webpack.config.js
```diff
@@ -17,7 +17,7 @@ module.exports = (env, options) => {
             filename: "taipy.js",
             path: path.resolve(__dirname, "../taipy/gui/webapp"),
             library: "Taipy",
-            publicPath: "/",
+            publicPath: "{{url_prefix}}",
```
FredLL-Avaiga: I don't think this will be replaced !
I guess we can remove this line completely

gui/src/utils/index.ts
```diff
@@ -99,9 +99,10 @@ export const getInitials = (value: string, max = 2): string =>
...
 
-export const ENDPOINT =
-    !process.env.NODE_ENV || process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test"
-        ? (process.env.REACT_APP_BACKEND_FLASK_URL as string)
-        : window.taipyUrl;
+export const getPathWithoutPrefix = (path: string, prefix: string): string => {
+    return path.startsWith(prefix) ? path.slice(prefix.length) : path;
+}
+
+export const ENDPOINT = window.taipyUrl;
```
FredLL-Avaiga, line 106: As discussed, can we remove this ?

taipy/gui/gui.py
```diff
@@ -408,9 +412,7 @@ def __send_ws(self, payload: dict) -> None:
         if grouping_message is None:
             try:
                 self._server._ws.emit(
-                    "message",
-                    payload,
-                    to=self.__get_ws_receiver(),
+                    "message", payload, to=self.__get_ws_receiver(), namespace=self._get_url_prefix() or "/"
```
FredLL-Avaiga, line 417: is url_prefix a namespace ?

gui/src/components/pages/MainPage.tsx
```diff
@@ -9,14 +10,15 @@ interface MainPageProps {
 }
 
 const MainPage = (props: MainPageProps) => {
+    const { state } = useContext(TaipyContext);
```
FabienLelaquais, line 13: I would certainly have used 'context' instead of 'state' (which has another meaning in the React world)

taipy/gui/config.py
```diff
@@ -37,7 +37,7 @@
 ]
 
 Config = t.TypedDict(
-    "AppConfig",
+    "Config",
```
FabienLelaquais, line 40: Hum sorry I missed that one... thanks

taipy/gui/gui.py
```diff
@@ -764,7 +771,7 @@ def add_pages(self, pages: t.Union[dict[str, t.Union[str, Page]], str] = None) -
 
                 if re_match := Gui.__RE_HTML.match(file_name):
                     renderers = Html(os.path.join(folder_path, file_name))
-                    renderers.modify_taipy_base_url(folder_name)
+                    renderers.modify_taipy_base_url(f"{folder_name}{self._get_url_prefix()}")
```
FabienLelaquais: Aren't the two fields reversed?

FabienLelaquais (approved):
Nice, nice job!

gui/src/components/pages/MainPage.tsx
FredLL-Avaiga: state and dispatch are the 2 attributes of context here and have been used throughout the front end so far

dinhlongviolin1: I dont think there is a problem here is it

taipy/gui/config.py
dinhlongviolin1: My pleasure :D

taipy/gui/gui.py
dinhlongviolin1: thanks, you are right

gui/webpack.config.js
dinhlongviolin1: it will be replaced by flask template. without this, the application will not work at all. Unless I am missing something here

gui/src/utils/index.ts
dinhlongviolin1: I think we can, but the base url on the frontend side will now be fully dependant on the current browser url. Like you mentioned yesterday, if the user deploys the application on apache for example and they have some custom route configuartion, our assumption that they will be using our routing configuration on the frontend might be incorrect in that case

taipy/gui/gui.py
dinhlongviolin1: it is a namespace since socket.io will now connect through /prefix

dinhlongviolin1:
> Why don't you use the same construct as in line 241?

You can only do it the way it is on line 241 because line 241 is an else while the below is an else if

dinhlongviolin1:
it will look pretty ugly like this `4 if a == 1 else (3 if a == 2 else (2 if a == 3 else 1))`

PR #111, opened 2022-03-17
table doc

FredLL-Avaiga:


gui/doc/chart.md
```diff
@@ -12,6 +12,12 @@ specify which trace you target.
 Indexed properties can have a default value (using the *property_name* syntax with
 no index) which is overridden by any specified indexed property.
 
+The _data_ property supported types are:
+- pandas Dataframe
```
FabienLelaquais, line 16: I suspect would give details on the behavior of the control at this point.
How are those different input types processed...

gui/doc/table.md
```diff
@@ -5,21 +5,58 @@ The table component supports 3 display modes:
    - _unpaginated_:  all rows and no page are shown (`show_all = True`).
    - _auto_loading_: the pages are loaded on demand depending on the visible area.
 
-## Details
+The _data_ property supported types are:
```
FabienLelaquais, line 8: I'm afraid you should give details on the behavior of the control at this point.
How are those different input types processed...

FabienLelaquais (approved):
Great job!

gui/doc/chart.md
FredLL-Avaiga: described in ### Multiple charts with different lengths

Issue #115, opened 2022-03-18
Remove client_url parameter


dinhlongviolin1:
**Description**
we don't need client_url parameter since the frontend side can use the browser endpoint to make request

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #101, opened 2022-03-14
Fix mypy errors


dinhlongviolin1:
**Description**
Make mypy happy. Possibly add mypy checks on github actions.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #117, opened 2022-03-18
remove client url + fix mypy error(#115, #101)

dinhlongviolin1:
Solve Issues #115, #101


gui/src/utils/index.ts
```diff
@@ -99,9 +98,6 @@ export const getInitials = (value: string, max = 2): string =>
         .join("")
         .toUpperCase();
 
-export const ENDPOINT =
-    !process.env.NODE_ENV || process.env.NODE_ENV === "development" || process.env.NODE_ENV === "test"
-        ? (process.env.REACT_APP_BACKEND_FLASK_URL as string)
-        : window.taipyUrl;
+export const ENDPOINT = window.location.origin;
```
FredLL-Avaiga: do we really need this ?
Could we not specify all our request without specifying the protocol://host:port part ?
Less is more :-)

FabienLelaquais (approved):
👌

gui/src/utils/index.ts
dinhlongviolin1: if we remove this, I dont know how to make a request :D

dinhlongviolin1: Do you suggest anything else Fred :D

FredLL-Avaiga: If I'm not mistaken any request can be written with "/path/to/resource" and the current protocol, host and port will be used.
even the io socket initialization would take window.location if no URL is specified ...

FredLL-Avaiga: so I'd say just remove ENDPOINT completely

dinhlongviolin1: Thanks Fred. I thought it wouldn't work in axios but it does after I tested it out so all good!

FredLL-Avaiga (approved):
🥇

Issue #122, opened 2022-03-21
BUG-Clicking on an already selected element in Toggle


FlorianJacta:
**Description**
When clicking twice on a toggle that is not the initial variable selected, nothing is selected (here click twice on Element 2). However, on the client, it seems that there is a selection made (here 'Initial element'). In other words, 'Initial element' seems to be selected on the client but is in reality not.

**How to reproduce**
```
from taipy.gui import Gui

status_list = ['Initial element', 'Element 2']
selected_status = None

md = """
# Select status
<|{selected_status}|toggle|lov={status_list}|>

**Selected status:** <|{selected_status}|>
"""

def on_change(state, var: str, val):
    print(var, type(val), val)


if __name__ == '__main__':
    Gui(page=md).run()
```

**Expected behavior**
Click on an unselect element => update value (ui elt selected)
Click on selected element => update value with unselected_value (ui selected not selected)

Issue #120, opened 2022-03-21
Partial: allow re-rendering with a different content


FredLL-Avaiga:
**Description**
Partial re-rendering would allow to change the content of a specific part of a page (contained in a part, dialog, pane or expandable)

This needs to be multi-user aware !

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #123, opened 2022-03-21
partial

FredLL-Avaiga:
- #120 add partial change of content
- fix partial not showing in part and expandable
- #122 toggle unselect

### DIFF
--- a/taipy/gui/gui_actions.py
+++ b/taipy/gui/gui_actions.py
@@ -103,7 +103,14 @@ def navigate(state: State, to: t.Optional[str] = ""):
     else:
         warnings.warn("'navigate()' must be called in the context of a callback")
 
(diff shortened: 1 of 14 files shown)


FabienLelaquais (approved):
I get it!
Nicely done

taipy/gui/gui_actions.py
```diff
@@ -3,6 +3,7 @@
 
 from .gui import Gui
 from .state import State
+from .partial import Partial
```
dinhlongviolin1: This import should be removed I guess

dinhlongviolin1 (approved):
Looks good to me! Simple yet effective solution for multi user

taipy/gui/gui_actions.py
FredLL-Avaiga: Good catch, tx

PR #132, opened 2022-03-22
Augment Reference Manual

FabienLelaquais:
Slight change in 'notify'


dinhlongviolin1 (approved):
Looks good to me

taipy/gui/partial.py
```diff
@@ -13,11 +13,17 @@ class Partial(_Page):
     and not related pages. This allows not to have to repeat yourself when
     creating your page templates.
 
-    Visual elements such as [`dialog`](../gui/viselements/dialog.md) or
+    Visual elements such as [`part`](../gui/viselements/part.md),
+    [`dialog`](../gui/viselements/dialog.md) or
```
FredLL-Avaiga, line 17: Are we missing expandable here?

FredLL-Avaiga (approved):
🐬

taipy/gui/partial.py
FabienLelaquais: I just say...'such as'. All elements are not listed here, and you are right.

FabienLelaquais: And I wanted to stress on part now that... it works.

PR #136, opened 2022-03-23
add state (#135)

FredLL-Avaiga:
- do not open browser when testing
- wait for table data to be loaded before checking


dinhlongviolin1 (approved):
Looks good to me

FabienLelaquais (approved):
... and I even understand it better now.
Thanks!

Issue #140, opened 2022-03-24
Configuration: env file settings should have a higher priority than Gui.run() options


FabienLelaquais:
**Description**

The configuration parameters should have a higher priority the simpler it is to set them.
As of today, taking the _port_ as an example:
- Say `taipy.gui.env` contains `port=1000`
- Say `Gui.run()` uses `port=1001`
- Say you run the script with `--port 1002`

What we expect is that Taipy uses the port 1002 (easier to set). And we do.

Now if you do NOT use that option, Taipy Gui uses port 1001 which I think is wrong: it is easier to change the env file than to change the code. So We expect '1000'.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #143, opened 2022-03-24
change config priority (#140)

dinhlongviolin1:
Solve Issue #140 

Now env file will have higher priority than gui.run() method


FredLL-Avaiga (approved):
👍

FabienLelaquais (approved):
Easy one, but better!

Issue #149, opened 2022-03-25
BUG-Display 'plotly' problem of charts when updated


FlorianJacta:
**Description**
When a chart is updated, texts or visual elements below goes behind the chart. When the chart is updated again, texts and visual elements will come back to their original places.

**How to reproduce**
Change `value` through the slider.
```
from taipy.gui import Gui
import pandas as pd

value = 100

dataset = pd.DataFrame({'x':[i for i in range(100)],
                        'y':[i for i in range(100)]})

displayed_dataset = dataset.copy()

md = """
<|{value}|slider|min=1|max=52|>

<|{displayed_dataset}|chart|x=x|y=y|>

TEXT OR OTHER ELEMENTS WILL GO BELOW THE GRAPH WHEN CHANGED
<|press|button|>
"""

def on_change(state, var_name: str, var_value):
    if var_name == 'value':
        state.displayed_dataset = dataset[:var_value]

Gui(md).run()
```

**Expected behavior**
Texts and visual elements should not change their places.

**Screenshots**
![pb_plotly](https://user-images.githubusercontent.com/98709993/160109675-1ea3621a-08e8-40ef-b6f7-127797a56a01.gif)

FredLL-Avaiga:
If we prevent the chart from being responsive, the problem disappears
```
from taipy.gui import Gui
import pandas as pd

pconf = {"responsive": False}
value = 100

dataset = pd.DataFrame({'x':[i for i in range(100)],
                        'y':[i for i in range(100)]})

displayed_dataset = dataset.copy()

md = """
<|{value}|slider|>

<|{displayed_dataset}|chart|plot_config={pconf}|>

TEXT OR OTHER ELEMENTS WILL GO BELOW THE GRAPH WHEN CHANGED
<|press|button|>
"""

def on_change(state, var_name: str, var_value):
    if var_name == 'value':
        state.displayed_dataset = dataset[:var_value]

Gui(md).run(use_reloader=False)
```
@FabienLelaquais Do you think we should:
- set responsive default to False
- add a responsive property to chart (default False)
- ...

FredLL-Avaiga:
OK then I'll change the default responsive to nothing and add some doc to explain how to make plotly charts responsive

FlorianJacta:
[...]

PR #150, opened 2022-03-25
#149 responsive chart

FredLL-Avaiga:
default set to plotly's default (False)

set reponsive via ```pconf = {"responsive": True}```
and ```<|{data}|chart|plot_config={pconf}|...|>```


FabienLelaquais (approved):
Mmmmouais

dinhlongviolin1 (approved):
Looks good to me

Issue #154, opened 2022-03-28
Gui.run: need a way to specify the frame in which to look for local variables


FredLL-Avaiga:
**Description**
In some specific cases, the user needs a way to specify where Gui looks for the variables used for bindings in the UI description.

Solution:
Best thing is to evaluate the frame from the Gui constructor and store it in Gui.__frame
add a _set_frame method to Gui to handle specific cases

This solution might be modified in the near future when we implement paged scopes

PR #158, opened 2022-03-29
#154 do it right

FredLL-Avaiga:
@florian-vuillemot I suspect you won't need Gui._set_frame anymore


dinhlongviolin1 (commented):
I am not sure about this. The __frame was previously the one that is in the run method

dinhlongviolin1 (approved):

FabienLelaquais (approved):
👍

florian-vuillemot (approved):
Wonderful!

PR #160, opened 2022-03-29
Expose the Gui class to the taipy package.

FabienLelaquais:
+ apply RefMan conventions.


taipy/__init__.py
```diff
@@ -1,12 +1,35 @@
-"""Top-level package for Taipy Gui.
+"""Taipy Graphical User Interface generator.
```
dinhlongviolin1: Correct me if i am wrong but shouldnt this file be placed in taipy/gui/__init__.py

FabienLelaquais: You are absolutely right.
This file should be more or less empty (except for the new import of the Gui class).
I'll fix that.
taipy/gui/__init__,py has been augmented properly.

dinhlongviolin1 (approved):
Looks good to me!

taipy/gui/__init__.py
```diff
@@ -1,15 +1,36 @@
...
+elements.
 
 Each page can contain regular text and images, as well
 as Taipy controls that are typically linked to some
 value that is managed by the whole Taipy application.
+
+Here is how you can create your first Taipy User Interface:
+
+   - Create a Python source file.
+     Copy these two lines into a file called _taipy.py_.
```
FredLL-Avaiga: no name clash with taipy
Even if not, I think it's not clear enough
Maybe use taipy_app.py

taipy/gui/gui.py
```diff
@@ -117,16 +130,16 @@ def __init__(
                 The default value is a file that has the same base name as the Python
                 file defining the `main` function, sitting next to this Python file,
                 with the `.css` extension.
-            path_mapping: TODO explain what this does.
-            env_filename: An optional file from which to load application configuration
-                variables (see the
+            path_mapping (Optional[dict]): TODO explain what this does.
```
FredLL-Avaiga: association between URL path and file system

taipy/gui/partial.py
```diff
@@ -43,8 +43,8 @@ def update_content(self, state: State, content: str):
         """Update partial content.
 
         Arguments:
-            state: the current user state as received in any callback.
-            content: the new content to show.
+            state: The current user state as received in any callback.
```
FredLL-Avaiga: no (State^) here ?

FredLL-Avaiga (approved):
a few comments

taipy/gui/__init__.py
FabienLelaquais: Makes sense.

taipy/gui/gui.py
FabienLelaquais: Let me give this a try.

taipy/gui/partial.py
FabienLelaquais: There should be.

Issue #171, opened 2022-04-01
BUG-update_content of a partial to change a table


FlorianJacta:
**Description**
On the example below, pressing the button should update my table (from data_1 to data_2). The column in common with the two dataframes is changing but not the one not in common. Clicking again will make it work.

**How to reproduce**
```
from taipy.gui import Gui
import pandas as pd

data_1 = pd.DataFrame({'toto_1':[1,2,3],'toto_2':[4,5,6]})
data_2 = pd.DataFrame({'titi_1':['carrot','egg','lemon'],'toto_2':['water','Cola','Limonade']})

data = data_1.copy()

md_for_partial = '<|{data}|table|width=50%|>'
main_md = '<|Change data|button|on_action=on_press|>\n<|part|partial={partial}|>'

def on_press(state):
    state.data = data_2.copy()
    state.partial.update_content(state, md_for_partial)

gui = Gui(main_md)
partial = gui.add_partial(md_for_partial)
gui.run()
```

**Expected behavior**
We should have the new data shown on the page (data_2) when pressing the button.

FredLL-Avaiga:
This is not linked to partial update but to the key identifying the data shown in the table

PR #174, opened 2022-04-01
update data key (#171)

FredLL-Avaiga:
this would allow to show to table based on the same data but showing different columns
#171


dinhlongviolin1 (approved):
Looks good to me

FabienLelaquais (approved):
Eyes closed.

Issue #181, opened 2022-04-04
BUG-Pb of compability of removeprefix for Multiple charts


FlorianJacta:
**Description**
Under versions older than Python 3.9, an error is raised when trying to create multiple charts when passing multiple data frames.

```
 UserWarning: Decoding Message has failed: {'type': 'DU', 'name': '_TpD_tp__data_1_data_2__0', 'payload': {'pagekey': '1/x_values-0/values-0/x_values-1/values', 'columns': ['1/x_values', '0/values', '0/x_values', '1/values'], 'alldata': True}, 'propagate': True, 'client_id': '20220303111220966695-0.04477817906557102'}
'str' object has no attribute 'removeprefix'
  warnings.warn(f"Decoding Message has failed: {message}\n{e}"
```

**How to reproduce**
With Python 3.8 installed, execute this code:
```
from taipy import Gui

import pandas as pd

data_1 = pd.DataFrame({'x_values':[1,2,3,4],'values':[23,14,45,18]})

data_2 = pd.DataFrame({'x_values':[3,4,5,6],'values':[42,12,35,26]})

Gui('<|{[data_1, data_2]}|chart|x=0/x_values|y[1]=0/values|x[1]=1/x_values|y[2]=1/values|>').run()
```

**Expected behavior**
We should get no error and the chart with the two traces from data_1 and data_2.

**Runtime environment**
Python 3.8.13

Issue #179, opened 2022-04-03
Improve data refresh


FredLL-Avaiga:
**Description**
When a backend action results in a change in a data variable (shown in table or chart), the controls have the responsibility of refreshing the data.
The current process has some "holes": if a data page is present in the front end but it's not displayed it is not invalidated resulting in potentially "un-fresh" data display.
We need to improve this behavior.

I propose to remove the data from the frontend when it is invalid.
This way, the controls will have to get a fresh page of data.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #182, opened 2022-04-04
refresh data fix bug

FredLL-Avaiga:
- #179 refresh data
- #181 removeprefix is python 3.9


taipy/gui/data/pandas_data_accessor.py
```diff
@@ -153,7 +153,7 @@ def __get_data(  # noqa: C901
     ) -> t.Dict[str, t.Any]:
         columns = payload.get("columns", [])
         if col_prefix:
-            columns = [c.removeprefix(col_prefix) for c in columns]
+            columns = [c[len(col_prefix):] for c in columns if c.startswith(col_prefix)]
```
FabienLelaquais: That is not the same semantics... are we good?

FabienLelaquais (approved):
👍

dinhlongviolin1 (approved):

taipy/gui/data/pandas_data_accessor.py
FredLL-Avaiga: No we're not. I thought of this last night

Issue #175, opened 2022-04-01
Support HTTPS for deployment (Heroku)


FlorianJacta:
**Description**
An error is now raised when trying to connect to the server from Heroku with HTTPS. The user has to force HTTP with Firefox. On other browsers, the user is not able to force HTTP.

![image](https://user-images.githubusercontent.com/98709993/161305428-498f72e4-8220-4a89-9879-c0d7b484e2ea.png)

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #184, opened 2022-04-04
support https (#175)

dinhlongviolin1:
Solve Issue #175 and return make get_flask_app public


taipy/gui/gui.py
```diff
@@ -1109,15 +1121,17 @@ def run(self, run_server: bool = True, run_in_thread: bool = False, **kwargs) ->
         self._bindings()._set_single_client(bool(app_config["single_client"]))
 
         # Start Flask Server
```
FredLL-Avaiga, line 1134: Need to update the return type of the run method and the doc ?

FredLL-Avaiga (approved):
need a bit of doc :-)

taipy/gui/gui.py
dinhlongviolin1: Thanks! I just update the types before seeing this message :D Will update the docs

FabienLelaquais (approved):
Looks just fine!

taipy/gui/gui.py
```diff
@@ -960,15 +966,15 @@ def _navigate(self, to: t.Optional[str] = ""):
...
+    def get_flask_app(self):
         return self._server.get_flask()
 
     def _set_frame(self, frame: FrameType):
         if not isinstance(frame, FrameType):
             raise RuntimeError("frame must be a FrameType where Gui can collect the local variables.")
         self.__frame = frame
 
-    def run(self, run_server: bool = True, run_in_thread: bool = False, **kwargs) -> None:
+    def run(self, run_server: bool = True, run_in_thread: bool = False, **kwargs) -> t.Optional[Flask]:
```
florian-vuillemot, line 986: Why Optional ?

florian-vuillemot (approved):

taipy/gui/gui.py
dinhlongviolin1: if run_server is True, it will return None

Issue #185, opened 2022-04-04
BUG-Markdown style support


FabienLelaquais:
**Description**

It looks like the Taipy parsing of Markdown does not allow Markdown inline styling.

The Markdown syntax should support the {} syntax as described below.

**How to reproduce**

If a Markdown page uses the following content:
```
# This is page 1 {.blue-class}

With some text in red. {#green-line}
```

And a CSS file would hold:
```
#green-line {
  color: rgb(0, 255, 17);
}

.blue-class {
  color: blue;
}
```

We expect the generated HTML to set the blue-class CSS class to the header, and give the paragraph the id 'green-line'.
The the header would be displayed in blue, and the paragraph, greenish.

FredLL-Avaiga:
for blocks the right markdown syntax is 
```
# This is page 1 {.blue-class}

With some text in red. 
{#green-line}
```

PR #186, opened 2022-04-04
style in markdown (#185)

FredLL-Avaiga:
and fix a tabular data column name

This allows to handle:
```
# This is page 1 {#idh1 .blue-class}
With some text in red. {test} { et encore { }
{#green-line .green-class}
```
and render it as (with color and style)
```
This is page 1
With some text in red. {test} { et encore { }
```


FabienLelaquais (approved):
👌

dinhlongviolin1 (approved):
nice one

Issue #188, opened 2022-04-05
add Taipy footer


FredLL-Avaiga:
**What would that feature address**
Show a taipy footer "Made with Taipy" or "Taipy inside"

***Description of the ideal solution***
this should be on by default but configurable by a config option ie change text or remove completely if option is None or empty string

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
Gui(md).run(use_reloader=False) => Taipy inside
Gui(md).run(use_reloader=False, watermark="Made by Fred")  => Made by Fred
Gui(md).run(use_reloader=False, watermark="") => no watermark

PR #193, opened 2022-04-05
#188 watermark

FredLL-Avaiga:
config via Gui.run(watermark="watermark")
text is centered at the bottom of the viewport with size = 0.7 em, color grey and opacity 0.4


FabienLelaquais (approved):
Neat!

dinhlongviolin1 (approved):

Issue #195, opened 2022-04-05
Theme cannot be set for light or dark, except with private APIs


FabienLelaquais:
**Description**

The theme[dark]/theme[light] configuration parameters are fine, expect they cannot be expressed as a parameter to Gui.run() or as an environment variable.

We must provide an alternative like 'dark_theme' & 'light_theme' that would have the same semantics as 'theme', but for their specific case.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
I propose that we change the name in the config to theme_dark & theme_light instead of theme[dark] ...
Would you agree @FabienLelaquais ?

PR #196, opened 2022-04-05
change config theme names (#195)

FredLL-Avaiga:


dinhlongviolin1 (approved):

taipy/gui/_default_config.py
```diff
@@ -12,8 +12,8 @@
     "favicon": None,
     "title": None,
     "theme": None,
-    "theme[light]": None,
-    "theme[dark]": None,
+    "theme_light": None,
```
FabienLelaquais: Why didn't you want to use, as the Issue stated, light_theme & dark_theme?

FabienLelaquais (approved):
Thanks.

taipy/gui/_default_config.py
FredLL-Avaiga: I can but I think that it's nice that they all start by theme*
Shall I change it ?

Issue #198, opened 2022-04-06
Performance problem on input control


FlorianJacta:
**Description**
The input control is calling the on_change function each time a character is added or deleted. It leads to a great number of calls and creates lags for this control.

You can normally see the lag in this demo: http://tp-production-planning.herokuapp.com/ (it has to be http, sometimes the server is in sleep and will take time to boot).

**Solution?**
A solution would be to call the on_change/update the state when the user has finished what he wanted to write. We could say that when the input is not selected anymore, it is time for the control to call the on_change and update the state.xxx of the control. We could say that this behavior is optional and is a parameter of the input control.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
use a timer to send the data only after 250ms if the user stopped typing

FlorianJacta:
Do you think having a buffer on sliders would also be great? Sometimes, I had some lag on sliders too when I update some other data when changing the value.

FredLL-Avaiga:
I'll look into it

FredLL-Avaiga:
I can implement a delay on slider but it's not very user friendly as the "nice" side of the slider is to see the impact on something in real time.
If you have performance issue with slider (I suppose because the data changed is big), you could activate the property _always_update_ (set to False) ie `don't always_update` so that the update is send only when you release the slider
The always_update_ property would make the n umber slider work like a LoV slider.
@FabienLelaquais advice ?

FabienLelaquais:
What about a global `update_latency` setting, that whose controls could override?
In a remote server situation, we would set that to 100-200 ms (250 look large to me...).
Locally, 0 would do just fine.

The idea is not to impact the UI code (the pages) to move from one situation to the other...
Would that be tricky to implement?

What are your thoughts?

PR #209, opened 2022-04-08
#198 user input buffer

FredLL-Avaiga:
apply a "rolling" timeout to the input so that we send the data only when the user has stopped typing for 300 ms


dinhlongviolin1 (approved):
I love it

FabienLelaquais (approved):
What else?

Issue #218, opened 2022-04-13
BUG-'dark_mode' parameter not working


FlorianJacta:
**Description**
The problem arises when trying to change the theme with the 'dark_mode' parameter of Gui().run(). The property doesn't seem to affect the theme of the client.

**How to reproduce**
Open an incognito tab to not have a default browser theme.
```
from taipy import Gui

Gui('# This is dark theme').run(dark_mode=False)
```

**Expected behavior**
We should be able to select the theme with the dark_mode parameter of Gui().run() and see the correct result on the client.

** Proposal **

- if the theme toggle is not present on the page, the theme is the one set by the backend (ignore local storage).
- if the theme toggle is present on the page, the theme is the one stored in the browser (persistent local storage) with default to the backend value.

FredLL-Avaiga:
proposal:
- if theme toggle is not present in the page, the theme is the one set by the backend (ignore local storage)
- if theme toggle is present in the page, the theme is the one stored in the browser (persistent local storage) with default to the backend value

PR #219, opened 2022-04-13
theme mode (#218)

FredLL-Avaiga:
- if theme toggle is not present in the page, the theme is the one set by the backend (ignore local storage)
- if theme toggle is present in the page, the theme is the one stored in the browser (persistent local storage) with default to the backend value


FabienLelaquais (approved):
That looks very reasonable, and far more easy to extend... Well done!

PR #224, opened 2022-04-14
Update deps

florian-vuillemot:


dinhlongviolin1 (commented):
Is there a reason for the fixed version in pipfile but not in setup.py

florian-vuillemot:
Pipfile: This allows the Taipy developer to work with the minimal API version of a package.
Setup.py: Minimum version that will be installed on the client computer. If the package is not installed, _pip_ will install the latest available and compatible.

This allows us to wear "ceinture et bretelle". During your development, you will test with the minimum version required for a client. If you need a new feature, you will have to update the _Pipfile_ to stick it with this new version. Of course, you should also fix the _setup.py_.. When you will push, the Github action (https://github.com/Avaiga/taipy-gui/blob/develop/.github/workflows/setuptools.yml) will run tests with the versions provided in the _setup.py_, so the latest.

FabienLelaquais (approved):
There's something I don't like about having devs using legacy versions. But I get the point.
We'll see how this goes.
As long as tests use the latest package versions... we should be good.

dinhlongviolin1 (approved):
Make sense to me

Issue #199, opened 2022-04-06
Associate an action to specific keys in an input field


FredLL-Avaiga:
**What would that feature address**
Press specific keys in an input control would fire the on_action 

***Description of the ideal solution***
- add an on_action property to input
- add a action_keys property list of authorized keys
  - default to Enter
  - supported keys: Enter, Escape, F1..F12
- on_action receives in payload.args [key, var_name, value]

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
Another way of dealing with special keys that would be more opened would be:
- add an on_action property to input
- add a action_keys property list of authorized keys 
  - default to Enter
  - supported keys: Enter, Escape, F1..F12
- on action receives in payload.args [key, var_name, value]

PR #225, opened 2022-04-15
#199 input on_action

FredLL-Avaiga:
- remove default on_action values


dinhlongviolin1 (approved):
Looks good to me.

Should we allow the users to specify the delay value for both on_input and on_action?

FabienLelaquais (approved):
Not a comma to change. Great!

Issue #227, opened 2022-04-19
BUG-'class_name' property of buttons not working


FlorianJacta:
**Description**
The class_name is not visible in the HTML of the page and cannot be used in the CSS file.

**How to reproduce**
```
from taipy.gui import Gui 

md = """
<|test 1|button|class_name=button_test|>
<|test 2|button|class_name=button_test|>
<|test 3|button|class_name=button_test|>
<|test 4|button|class_name=button_test|>
"""

Gui(md).run()
```

**Expected behavior**
We should be able to see the class in the HTML page and interact with it in the CSS file.

**Runtime environment**
taipy 1.0.0

Issue #226, opened 2022-04-19
BUG-Time zone configuration and tables


FlorianJacta:
**Description**
Two series from different time zones give the same result (normal from your explanation). However, it gives the UTC time even when the time zone is changed in the configuration. The Time zone configuration doesn't seem to have an impact on the table.

**How to reproduce**
```
from taipy.gui import Gui, Markdown
import pandas as pd
import numpy as np
import pytz

japan_tz = pytz.timezone('Asia/Tokyo')
df = pd.DataFrame({'Japan time zone': pd.date_range(start='2022-01-01', periods=100, tz=japan_tz), 'y': np.random.randint(100, 200, size=100)})
df.loc[:, 'UTC time zone'] = df.loc[:, 'Japan time zone'].dt.tz_localize(None).dt.tz_localize('UTC')
df.loc[:, 'Japan time zone'] = df.loc[:, 'Japan time zone']
md = """
# Datetime in table

<|{df}|table|width=100%|>
"""

print(df)
Gui(md).run(time_zone='Asia/Tokyo')
```

**Expected behavior**
We should get the correct time considering we are in the Japan Time zone.

**Screenshots**
![image](https://user-images.githubusercontent.com/98709993/163957004-543575cc-06a9-43c8-8be9-95ff62e142cc.png)

**Runtime environment**
taipy: 1.0.0

PR #228, opened 2022-04-19
timezone and class_name (#226 #227)

FredLL-Avaiga:
- #226 time zone configuration
  - convert dates to UTC before transmitting
  - use dataframe column timezone if present
  - client timezone defaults to client with column use if not set
- #227 property classname => class_name


dinhlongviolin1 (approved):
Was about to ask a few things but I answered it myself after reading through it a few times

FabienLelaquais (approved):
Looks good to me

PR #229, opened 2022-04-20
New release and publish process

florian-vuillemot:


dinhlongviolin1:
With the new release process, the version would have to be provided in the input of workflow_dispatch right

florian-vuillemot:
Yes. The version should be the same than in the setup.py + major & minor should match the branch name ("release/Major.Minor")

florian-vuillemot:
@FabienLelaquais @dinhlongviolin1 important point. The release does not build the web app anymore. So, if a user installs taipy from a Github release, npm has to be installed.
The behaviour of Pypi is not the same because Pypi wants a package already built. And the web app generation is done by the build.

FabienLelaquais (approved):
Looks good to me

FabienLelaquais:
> @florian-vuillemot I'm not sure I understand the npm build part of this though.
A release must contain the WebApp, and users should never have to have npm installed locally.
Are we on the same page?

.github/workflows/release.yml
```diff
@@ -1,81 +1,79 @@
...
+          rm -rf taipy
+          pip install dist/*.tar.gz
+          pip install pytest pytest-mock
+          pytest
 
-    runs-on: ${{ matrix.os }}
+  release:
+    needs: [test-package]
+    timeout-minutes: 20
+    name: Release taipy-core Package
```
dinhlongviolin1: taipy-gui

dinhlongviolin1 (approved):
Looks solid to me

.github/workflows/release.yml
florian-vuillemot:
```suggestion
    name: Release package
```

Issue #234, opened 2022-04-25
BUG-Alert user when the port is already used


FredLL-Avaiga:
**Description**
I you start to Taipy python scripts with the same port, no error message is displayed although only the first server addresses the requests

**How to reproduce**
- start taipy python script 1
- start taipy python script 2

Access the taipy main page (localhost:5000), the first sevrer answers

**Expected behavior**
The second server shouldn't start

FredLL-Avaiga:
pushed to release/1.0

PR #235, opened 2022-04-25
#234 check port before starting

FredLL-Avaiga:


dinhlongviolin1 (approved):
Nice one

dinhlongviolin1 (approved):
Nicely done

FredLL-Avaiga:
The exception is falsely raised when use_reloader is True and the reloader kicks in :-(
Any idea @dinhlongviolin1 ?

dinhlongviolin1:
> The exception is falsely raised when use_reloader is True and the reloader kicks in :-(
> Any idea @dinhlongviolin1 ?

I have no idea. But if it is because of the reloader, you could use the `is_running_from_reloader` guard to check if it is running with a reloader and only perform the port check outside of the reloader.

taipy/gui/server.py
```diff
@@ -230,6 +231,13 @@ def _run_notebook(self):
 
     def runWithWS(self, host, port, debug, use_reloader, flask_log, run_in_thread, ssl_context):
         host_value = host if host != "0.0.0.0" else "localhost"
+        if debug and not is_running_from_reloader():
+            # Check that the port is not already opened
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            result = sock.connect_ex((host_value,port))
```
FabienLelaquais, line 237: No complaints from black with the space missing after the comma?

FabienLelaquais (approved):
This should go into 1.0 as well I'm afraid

taipy/gui/server.py
FredLL-Avaiga: nope

Issue #236, opened 2022-04-25
BUG-Partials not updated when updated by a blank string


FlorianJacta:
**Description**
state.partial.update_content(page) doesn't seem to update the page when the page is empty. It is therefore impossible to update the partial to a "blank partial".

**How to reproduce**
```
""Partials and empty strings.

To reproduce:
1. Type something in the input box
2. Ctrl+A to select all the text in the input box
3. Delete

The partial doesn't update.
"""

from taipy.gui import Gui, Markdown

text = ""

md = """
<|toggle|theme|>

# Testing partials

## Enter value:

<|{text}|input|on_change=update_partial|>

## Real text:

<|{repr(text)}|>

## Partial content:

<|part|partial={partial}|>

"""


def update_partial(state, var_name: str, var_value):
    state.partial.update_content(state, state.text)


if __name__ == "__main__":
    gui = Gui(page=Markdown(md))
    partial = gui.add_partial("")
    gui.run(dark_mode=False, port=5001)
```

**Expected behavior**
In this example, when we delete texts in the input, the partial is not properly updated. When the input is blank "", we should get nothing in the partial instead of a letter or word. We should be able to update to a "blank partial".

FredLL-Avaiga:
pushed to release/1.0

PR #237, opened 2022-04-25
#236 partial update with empty string

FredLL-Avaiga:


dinhlongviolin1 (approved):
💯

FabienLelaquais (approved):
Makes senses. Should make its way in 1.0.3...

FredLL-Avaiga:
> Makes senses. Should make its way in 1.0.3...

Tx. It will

Issue #153, opened 2022-03-28
Initiate the Maps integration


FabienLelaquais:
See what we can do with Maps from Plotly.

FredLL-Avaiga:
plotly.js
 examples: https://plotly.com/javascript/mapbox-layers/

FredLL-Avaiga:
We could also provide a Gmaps specific control using:
- [a react wrapper](https://www.npmjs.com/package/google-map-react)
- [clusters](https://www.leighhalliday.com/google-maps-clustering)
or more basically
- [from Google Dev](https://developers.google.com/maps/documentation/javascript/react-map)

FredLL-Avaiga:
1st step: support lat/lon as property of chart

FredLL-Avaiga:
```
from taipy.gui import Gui
import pandas as pd

data = pd.read_csv("https://raw.githubusercontent.com/plotly/datasets/master/2015_06_30_precipitation.csv")

marker = { "color": "fuchsia", "size": 4 }

layout = {
			"dragmode": "zoom",
			"mapbox": { "style": "open-street-map", "center": { "lat": 38, "lon": -90 }, "zoom": 3 },
			"margin": { "r": 0, "t": 0, "b": 0, "l": 0 }
		}

md = """
# Test Map
<|{data}|chart|type=scattermapbox|marker={marker}|layout={layout}|lat=Lat|lon=Lon|text=Globvalue|mode=markers|>
"""

Gui(md).run(use_reloader=False)
```

PR #239, opened 2022-04-27
support lat lon (#153)

FredLL-Avaiga:
1st step
I think I'll create another control specific to GMaps

### DIFF
--- a/gui/doc/chart.csv
+++ b/gui/doc/chart.csv
@@ -4,9 +4,11 @@ type,indexed(str),scatter,"Chart type.
 See the Plotly <a href=""https://plotly.com/javascript/reference/"">chart type</a> documentation for details."
 mode,indexed(str),lines+markers,"Chart mode.
 See the Plotly <a href=""https://plotly.com/javascript/reference/scatter/#scatter-mode"">chart mode</a> documentation for details."
-x,indexed(str),,Column name for the <i>x</i> axis.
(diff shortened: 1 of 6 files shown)


dinhlongviolin1 (approved):
Looks simple and works like a charm

gui/doc/chart.csv
```diff
@@ -4,9 +4,11 @@ type,indexed(str),scatter,"Chart type.
 See the Plotly <a href=""https://plotly.com/javascript/reference/"">chart type</a> documentation for details."
 mode,indexed(str),lines+markers,"Chart mode.
 See the Plotly <a href=""https://plotly.com/javascript/reference/scatter/#scatter-mode"">chart mode</a> documentation for details."
-x,indexed(str),,Column name for the <i>x</i> axis.
-y,indexed(str),,Column name for the <i>y</i> axis.
+x,indexed(str),,Column name for the <i>x</i> axis. Synonym <i>lat</i>.
```
FabienLelaquais, line 7: That's is quite strange.
Latitude indicates how much north or south you are (indicating a 'y' notion). And Longitude is more 'x'ish...

Rephrase: "Synonym for <i>*</i>."

FabienLelaquais (approved):
Looks good

gui/doc/chart.csv
FredLL-Avaiga: Didn't realize that :-)
I'll swap them if it makes more sense

Issue #238, opened 2022-04-26
Control like an input for password


FlorianJacta:
**What would that feature address**
Now it is not possible to have an input for password so without auto complete and without seeing the password. 

***Description of the ideal solution***
The goal will be to have a similar control as the input control with type="password" that will allow the user to write its password without seeing the actual input

***Caveats***
It is a new control.

***Other options***
Adding type="password" in input control.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
As a first step I'll add a property password to the field control.
In the future, we might to find a way to encode the input data ...

PR #245, opened 2022-04-28
#238 password

FredLL-Avaiga:
adding password boolean property to the input control

### DIFF
--- a/gui/doc/input.csv
+++ b/gui/doc/input.csv
@@ -1,6 +1,6 @@
 ﻿Name,Type,Default,Doc
 *value,dynamic(any),None,The value represented by this control.
-password,bool,False,"Sets the input as a password (ie character are displayed as *)."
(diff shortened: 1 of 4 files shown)


gui/doc/input.csv
```diff
@@ -1,5 +1,6 @@
 ﻿Name,Type,Default,Doc
 *value,dynamic(any),None,The value represented by this control.
+password,bool,False,"Sets the input as a password (ie character are displayed as *)."
```
FabienLelaquais: -> "If True, the text is obscured: all input characters are displayed as an asterisk ('*')."

Note for Fred: if this is not dynamic, we cannot create any 'Show password' toggle button, right? Just sayin'

FabienLelaquais (approved):
Nice

gui/doc/input.csv
FredLL-Avaiga: I guess this would never need to be dynamic.
We could add an option to allow the user to reveal the password but it would be anoter property and the reveal would be local

PR #253, opened 2022-05-04
Small doc fixes based on incoming comments.

FabienLelaquais:


FabienLelaquais:
Does not seem to be based on release/1.0, which is the objective here.

Issue #257, opened 2022-05-05
Prevent the same variable to be evaluated multiple times in on_change function


dinhlongviolin1:
**Description**
On the current version, if the user try to update a variable on the `on_change` method which has already been updated before, python will crash due to inifite loop leading to stackoverflow. The below code demonstate how to reproduce the issue.

```
from taipy.gui import Gui

x = 10

def on_change(state, var, val):
    if var == "x":
        state.x = val * 2

gui = Gui(page="""

x = <|{x}|>

x * 10 = <|{x*10}|>

Slider for x: <|{x}|slider|>

""")

gui.run()
```

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #258, opened 2022-05-05
Prevent the same variable to be evaluated multiple times in on_change function (#257)

dinhlongviolin1:
Solve Issue #257 

With this udpate, variable x in the on_change function wont be updated in this case since x has been updated before --> No inifinite loop.

```
from taipy.gui import Gui

x = 10

def on_change(state, var, val):
    if var == "x":
        state.x = val * 2

gui = Gui(page="""

x = <|{x}|>

x * 10 = <|{x*10}|>

Slider for x: <|{x}|slider|>

""")

gui.run()
```

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -190,7 +190,7 @@ class Gui:
         self.on_change: t.Optional[t.Callable] = None
         self.on_init: t.Optional[t.Callable] = None
 
-        # gui syncrhonous state variable
+        # gui synchronous state variable
         self.__modified_vars_update_var: t.Optional[t.Set] = None
 
         # Load default config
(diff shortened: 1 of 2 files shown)


taipy/gui/gui.py
```diff
@@ -190,6 +190,9 @@ def __init__(
         self.on_change: t.Optional[t.Callable] = None
         self.on_init: t.Optional[t.Callable] = None
 
+        # gui syncrhonous state variable
```
FredLL-Avaiga: typo in synchronous

FredLL-Avaiga (approved):
Well done

taipy/gui/gui.py
dinhlongviolin1: thanks

FabienLelaquais (approved):
Simple enough. Good job!

Issue #259, opened 2022-05-09
BUG-CSS file not loaded


dinhlongviolin1:
**Description**
During deployment, the default css file name is fetched from the filename of the main module. The __main__ module will not be the one that the user test their code on.

**Acceptance Criteria**
The default css file will be loaded correctly

PR #260, opened 2022-05-09
fix default css filename (#259)

dinhlongviolin1:
Fix bug #259 cssfile name for FJ

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -25,8 +25,8 @@ from types import FrameType
 
 import __main__
 import markdown as md_lib
-from flask import Blueprint, Flask, request, send_from_directory
 import tzlocal
+from flask import Blueprint, Flask, request, send_from_directory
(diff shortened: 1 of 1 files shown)


taipy/gui/gui.py
```diff
@@ -25,8 +25,8 @@
 
 import __main__
 import markdown as md_lib
-from flask import Blueprint, Flask, request, send_from_directory
 import tzlocal
+from flask import Blueprint, Flask, request, send_from_directory
```
FabienLelaquais, line 29: How come these were added?

FabienLelaquais (approved):
👍

FredLL-Avaiga (approved):
+1

taipy/gui/gui.py
dinhlongviolin1: These were here from the beginning. Pre-commit just formatted it and move it to another line

PR #262, opened 2022-05-14
upd: black version on pre-commit

joaoandre-avaiga:
Hello!

Im creating a PR in all repos to update the version of black. The version we are using is very outdated and it started to fail because of dependency issues.


dinhlongviolin1 (approved):
I have done this in another PR actually but it is better to merge it now

Issue #84, opened 2022-03-07
Make the creation of the layout columns easier


FlorianJacta:
**Description**

This improvement will add to the existing syntax, another way to create columns in the layout parameters. Currently, when we want to create 5 columns in a layout, we write
```
<|layout|columns=1 1 1 1 1
```
The goal would be to change this syntax to make it simpler and easier to read. For example, like this :
```
<|layout|columns=1*5
```
This syntax will create the same layout as before so 5 columns of the same width.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
Do you want to also support 2\*2+3\*1 ?

FlorianJacta:
Well, if it is possible. Why not

FredLL-Avaiga:
Not sure it's clearer that way
The thing is that we support classic css grid column description  
https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns

PR #273, opened 2022-05-20
Allow for column expression as 1*5 (#84)

FredLL-Avaiga:


gui/src/components/Taipy/Layout.tsx
```diff
@@ -16,12 +16,29 @@ const baseSx = { display: "grid" };
 
 const HAS_TEXT = /[^\d\s]/;
 const SPLIT_COLS = /\D/;
+const EXPR_COLS = /\D*(\d)+[^\d*]*\*[^\d*]*(\d)+\D*/;
+
+const expandCols = (cols: string) => {
```
FabienLelaquais, line 18: Nice way of dealing with that!

Aren't we losing potentially important info here? And isn't that regexp too lazy ("3foo*bar1" will match)?
And I'm not sure that "2*1 3*2" will work either...

FabienLelaquais (changes requested):
Nicely done!
Can you please look at the last question of my single comment?

gui/src/components/Taipy/Layout.tsx
FredLL-Avaiga: 3foobar1 will not match as we request a * in between

FredLL-Avaiga: 21 3*2 will be treated as 3\*2 which is fine I think

FredLL-Avaiga:
> Nicely done! Can you please look at the last question of my single comment?

I did ;-)

gui/src/components/Taipy/Layout.tsx
FredLL-Avaiga: @FabienLelaquais Are you happy with this ?

FabienLelaquais: Yeah my comment on 3boobar1 looks wrong, because the star symbol is consumed by the MD formatting..

However: in my opinion, "2 1 3\*2" should be equivalent to "2 1 2 2 2" which I find will not work, since the match is performed only once. Same for "2\*1 3\*2" for example (creating "1 1 2 2 2")... Am I missing something?

FredLL-Avaiga: I get your point
Will do

FredLL-Avaiga: I tried to be more clever:
- 1 1 1 1 => 1fr 1fr 1fr 1fr
- 10px 1em 1rem 1fr => 10px 1em 1rem 1fr
- 1\*3px 1 3\*1em 2 3 3\*2 => 3px 1fr 1em 1em 1em 2fr 3fr 2fr 2fr 2fr

FabienLelaquais (approved):
Great!!

dinhlongviolin1 (approved):

PR #278, opened 2022-05-28
gui auth

dinhlongviolin1:
Code for auth features on Frontend

### DIFF
--- a/gui/src/context/taipyReducers.ts
+++ b/gui/src/context/taipyReducers.ts
@@ -27,6 +27,8 @@ enum Types {
     SetMenu = "SET_MENU",
     DownloadFile = "DOWNLOAD_FILE",
     Partial = "PARTIAL",
+    SetToken = "SET_TOKEN",
+    RemoveToken = "REMOVE_TOKEN",
 }
 
 export interface TaipyState {
@@ -44,6 +46,7 @@ export interface TaipyState {
     id: string;
     menu: MenuProps;
     download?: FileDownloadProps;
+    token?: string;
 }
 
 export interface TaipyBaseAction {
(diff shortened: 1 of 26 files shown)


gui/src/context/taipyReducers.ts
```diff
@@ -163,6 +170,7 @@ export const INITIAL_STATE: TaipyState = {
         : undefined,
     id: getLocalStorageValue("TaipyClientId", ""),
     menu: {},
+    token: localStorage && localStorage.getItem("token") || undefined,
```
FredLL-Avaiga: Why not use getLocalStorageValue ?

taipy/gui/gui.py
```diff
@@ -1033,7 +1033,17 @@ def _set_frame(self, frame: FrameType):
             raise RuntimeError("frame must be a FrameType where Gui can collect the local variables.")
         self.__frame = frame
 
-    def run(self, run_server: bool = True, run_in_thread: bool = False, ssl_context: t.Optional[t.Union[ssl.SSLContext, t.Tuple[str, t.Optional[str]], t.Literal['adhoc']]] = None, **kwargs) -> t.Optional[Flask]:
+    def _set_state(self, state: State):
```
FredLL-Avaiga, line 1080: I suppose this is used by gui enterprise ?

FredLL-Avaiga (approved):
need to review enterprise

gui/src/context/taipyReducers.ts
dinhlongviolin1: You are right thanks!

taipy/gui/gui.py
dinhlongviolin1: Yes you are correct! The new state class is used by the enterprise

FabienLelaquais (approved):
👍

Issue #248, opened 2022-04-28
Add callbacks for a state


FlorianJacta:
**What would that feature address**
It is not possible to give a callback function to change the state of one user or for all the users. For example, it will allow the user to add a callback function to Taipy Core jobs and scenarios. Like this, the user's state or all the states "can be warned" of any changes in the backend through this callback.

***Description of the ideal solution***
The ideal solution will be to provide a way to create this callback either to callback a specific state or all the states. It should be compatible with the backend callbacks (subscribers of jobs/scenarios). It will also be ideal to stay close to the format of the other Gui callbacks (on_change/on_action...).

***Caveats***

***Other options***

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
We need to:
- add a Gui action to get a context id that can be provided to a backend callback (ie serializable)
- add a Gui action to call a user callback with the State given a context id

@FabienLelaquais @dinhlongviolin1 What are your thoughts ?

FredLL-Avaiga:
This version is not directly serializable, one has to use a workaround with Taipy-core:

```py3
val = 1

def user_callback(state: State, *args):
    print(f"In user_callback(state, {args})")
    state.val = state.val * 10

suscribe_cb = None

def on_serializable_callback():
    if callable(suscribe_cb):
        suscribe_cb("from serializable_callback")

def on_init(state: State):
    suscribe_cb = state._gui.get_callback(user_callback)
    subscribe(on_serializable_callback) # from taipy-core
```

FredLL-Avaiga:
the user code would look like:

```
from taipy.Gui import Gui, State, get_context_id, invoke_state_callback

val = 1

aGui = Gui(some_md)

def a_user_callback(state: State, scenario, job):
    state.val = 10
    notify(state, "I", f"scenario {scenario} is done")

def a_serializable_callback(scenario, job, context_id: str):
    invoke_state_callback(aGui, context_id, user_callback, scenario, job)

def on_init(state: State):
    scenario.subscribe(a_serializable_callback, {"context_id": get_context_id(state)})

aGui.run()
```

PR #279, opened 2022-05-30
manage context id to prepare for serializable callback (#248)

FredLL-Avaiga:
If one wants to use this with a serializable function: 

```
val = 1

def user_callback(state: State, *args):
    print(f"In user_callback(state, {args})")
    state.val = state.val * 10

suscribe_cb = None

def on_serializable_callback():
    if callable(suscribe_cb):
        suscribe_cb("from serializable_callback")

def on_init(state: State):
    global suscribe_cb
    suscribe_cb = state._gui.get_callback(user_callback)
    subscribe(on_serializable_callback) # from taipy-core
```


dinhlongviolin1 (approved):
Looks all good!

Thanks for the addition of session id 

nice changes to allow set client id from multiple places by the reintroduction of adding the client id to request

FabienLelaquais (approved):
This I get.

Issue #289, opened 2022-06-10
Add the 'label' property to the selector control


FlorianJacta:
**Description**
It would add the 'label' property to the selector.

![label](https://user-images.githubusercontent.com/98709993/173102156-57c6931e-8eff-41ae-ba98-c902b99592a8.gif)

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #292, opened 2022-06-13
label property for input, number and dropdown selector (#289)

FredLL-Avaiga:


dinhlongviolin1 (approved):

gui/src/components/Taipy/Selector.tsx
```diff
@@ -181,12 +210,14 @@ const Selector = (props: SelTreeProps) => {
...
         <FormControl sx={controlSx} className={className}>
+            {props.label ? <InputLabel>{props.label}</InputLabel> : null}
             <Tooltip title={hover || ""}>
                 <Select
+                    id={id}
                     multiple={multiple}
                     value={dropdownValue}
                     onChange={handleChange}
-                    input={<OutlinedInput />}
+                    input={<OutlinedInput label={props.label} />}
```
FabienLelaquais, line 220: No issue here if props.label is null?

FabienLelaquais (approved):
👍

gui/src/components/Taipy/Selector.tsx
FredLL-Avaiga: Nope (undefined)

Issue #286, opened 2022-06-09
BUG- Cannot change the State of a variable twice in the same block


FlorianJacta:
**Description**
In Taipy-Gui 1.1.0, it is not possible to change the state of a variable twice or more. The new value would not be assigned. On the code below, the first `state.ffilter` will be well assigned but not the second.

**How to reproduce**
```python
from taipy.gui import Gui, State
import pandas as pd
import typing as t

sm_dayofweek = [('0','Mon'),('1','Tue'),('2','Wed'),('3','Thu'),('4','Fri'),('5','Sat'),('6','Sun')]
sm_day_selected = None

data = pd.DataFrame({"Col 1": [1, 2, 3], "Col 2": [10, 20, 30], "Col 3": [100, 200, 300]})
ffilter = [True, True, True]

def change_days(state: State, var, val):
    print(f"change_days(state: State, var: {var}, val: {val})")
    state.ffilter = pd.Series([True, True, True])
    new_filter:t.List[bool] = [False, False, False]
    for i, v in enumerate(state.ffilter):
        new_filter[i] = str(i) in val
    state.ffilter = pd.Series(new_filter)
    print(pd.Series(new_filter))
    print(state.ffilter)

md = """
<|{sm_day_selected}|selector|lov={sm_dayofweek}|on_change=change_days|dropdown|multiple|value_by_id|>
<|{data[ffilter]}|chart|>
"""

Gui(md).run(use_reloader=False)
```

**Expected behavior**
We should be able to change the state of a variable twice or more.

**Runtime environment**
Please specify relevant indications.
 - OS:Windows
 - Taipy-Gui 1.1.0

FredLL-Avaiga:
issue is linked to the protection to not enter in an infinite loop on update
`gui.py:344`

I suppose `self.__modified_vars_update_var` is not thread safe

PR #293, opened 2022-06-14
Change var twice in on_change (#286)

FredLL-Avaiga:
State variable will be updated but the on_change will only be called the first time

### DIFF
--- a/taipy/gui/gui.py
+++ b/taipy/gui/gui.py
@@ -284,7 +284,7 @@ class Gui:
         if der_vars is None:
             g.derived_vars = derived_vars
         else:
-            der_vars.update(derived_vars),
+            der_vars.update(derived_vars)
         if var_name in modified_vars:
             return True
         else:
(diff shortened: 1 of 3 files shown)


taipy/gui/gui.py
```diff
@@ -272,10 +269,41 @@ def __set_client_id_in_context(self, client_id: t.Optional[str] = None):
...
+            der_vars.update(derived_vars),
+        if var_name in modified_vars:
+            return True
+        else:
+            modified_vars.add(var_name)
+            return False
+
+    def __clean_vars_on_exit(self) -> t.Optional[t.Set[str]]:
+        update_count = getattr(g, "update_count", 0) - 1
+        print(f"__clean_vars_on_exit: {update_count} \n\tmodified_vars: {getattr(g, 'modified_vars', None)} \n\tderived_vars: {getattr(g, 'derived_vars', None)}")
```
FabienLelaquais: Are you sure you want to keep this??

FabienLelaquais (approved):
I... think I got it.

taipy/gui/gui.py
FredLL-Avaiga: oups 👎

dinhlongviolin1 (approved):
I knew this would need to be dealt with as soon as we use g. Well done Fred!

Issue #297, opened 2022-06-16
BUG-Running on Watson studio


vhfgosselin:
**Description**
TAipy GUI no able to execute on watson studio

**How to reproduce**
```
from taipy.gui import Gui, Markdown

page = Markdown("""
# Taipy in a Notebook

Value: <|{value}|>

Set: <|{value}|slider|>
""")

value = 10

gui = Gui(page)
gui.run()
```
leads to:
AttributeError                            Traceback (most recent call last)
/tmp/wsuser/ipykernel_364/2190609130.py in <module>
     11 value = 10
     12 
---> 13 gui = Gui(page)
     14 gui.run()

/opt/conda/envs/Python-3.9/lib/python3.9/site-packages/taipy/gui/gui.py in __init__(self, page, pages, css_file, path_mapping, env_filename, flask)
    174         # store suspected local containing frame
    175         self.__frame = t.cast(FrameType, t.cast(FrameType, inspect.currentframe()).f_back)
--> 176         self.__default_module_name = _get_module_name_from_frame(self.__frame)
    177 
    178         # Preserve server config for server initialization

/opt/conda/envs/Python-3.9/lib/python3.9/site-packages/taipy/gui/utils/get_module_name.py in _get_module_name_from_frame(frame)
     17 
     18 def _get_module_name_from_frame(frame: FrameType):
---> 19     return t.cast(ModuleType, inspect.getmodule(frame)).__name__
     20 
     21 

AttributeError: 'NoneType' object has no attribute '__name__'

**Expected behavior**
Description of what would be the expected outcome.

**Screenshots**
![image](https://user-images.githubusercontent.com/20767690/174115590-01be5ca4-8131-44db-acf0-211491834a34.png)

**Runtime environment**
- Watson Studio
- Taipy 1.1

PR #303, opened 2022-06-22
get module_name through f_globals (#297)

dinhlongviolin1:
Fix Bug #297


FredLL-Avaiga (approved):
Makes sense I suppose
Is it an issue with Notebooks or DrWatson ?

dinhlongviolin1:
> Makes sense I suppose Is it an issue with Notebooks or DrWatson ?

It is an issue with both. Now both is running properly!

FabienLelaquais (approved):
If you say this works... this works.
Great!

Issue #173, opened 2022-04-01
Update to React 18


FredLL-Avaiga:
**Description**
Stay current: Update to React 18

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
too early for dependencies

FredLL-Avaiga:
if you want to build locally:
- install  with npm install --omit=optional
(you might want to remove node_modules and/or  to start efresh)

PR #308, opened 2022-06-23
upgrade react to v18 (#173)

FredLL-Avaiga:
- upgrade test libraries
- install without optional dependencies (npm install --omit=optional) to prevent wront version of react types to be installed
- install @types/react
- specify missing types
- remove deprecation warnings


dinhlongviolin1 (approved):
all works just fine. well done Fred!

gui/src/components/Taipy/DateSelector.spec.tsx
```diff
@@ -1,9 +1,9 @@
 import React from "react";
-import { render, fireEvent } from "@testing-library/react";
+import { render } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import "@testing-library/jest-dom";
 
-import { LocalizationProvider } from "@mui/lab";
+import { LocalizationProvider } from "@mui/x-date-pickers";
```
FabienLelaquais, line 6: Hmmm... Introduction of Mui X...

setup.py
```diff
@@ -50,7 +50,7 @@
 def _build_webapp():
     already_exists = Path(f"./taipy/gui/webapp/index.html").exists()
     if not already_exists:
-        os.system("cd gui && npm ci && npm run build")
+        os.system("cd gui && npm ci --omit=optional && npm run build")
```
FabienLelaquais, line 53: Shouldn't there be a --only=prod at this point?

FabienLelaquais (approved):
Great, great job!

setup.py
FredLL-Avaiga: --only=prod is not supported since at least npm 6.3

gui/src/components/Taipy/DateSelector.spec.tsx
FredLL-Avaiga: it replaces @mui/lab for date pickers and I'd say it is more stable than lab

PR #309, opened 2022-06-24
fix import all error (#310)

dinhlongviolin1:
bug Avaiga/taipy-gui#310


FredLL-Avaiga (approved):
easy enough

FredLL-Avaiga (approved):
Should we change the version back to 1.1.... ?

FredLL-Avaiga (approved):
I will create a separate PR for the 1.1

FabienLelaquais (approved):
👍

Issue #311, opened 2022-06-24
Use gevent-websocket instead of simple-websocket


dinhlongviolin1:
**Description**

Taipy needs to support Conda but `simple-websocket` is not available in Conda. We need to switch to `gevent`.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #312, opened 2022-06-24
gevent-websocket (#311)

dinhlongviolin1:
Solve Issue #311


tox.ini
```diff
@@ -28,6 +28,7 @@ commands =
 platform = linux
 commands =
     pipenv install --dev --skip-lock
+    pipenv run pip uninstall gevent gevent-websocket -y
```
FredLL-Avaiga, line 31: I thought we could use pipenv uninstall gevent ... (without run pip) ?

FredLL-Avaiga (approved):

tox.ini
dinhlongviolin1: it is equivalent in this case since we won't be using piplock

taipy/gui/gui.py
```diff
@@ -1270,6 +1271,11 @@ def run(
                 Configure TLS to serve over HTTPS. Can be an ssl.SSLContext object, a (cert_file, key_file) tuple to
                 create a typical context, or the string 'adhoc' to generate a temporary self-signed certificate.</br>
                 The default value is None.
+            async_mode (Optional[str]): A configuration of Flask-SocketIO. Valid async modes are `threading`, `eventlet`,
```
FabienLelaquais: I suspect we are going to need more explanations on this one.
Let's talk about it.

FabienLelaquais (approved):
We will need a bit more doc on this Long.
Good work though!

Issue #315, opened 2022-06-28
Allow to access module name from state class


dinhlongviolin1:
**Description**
With this feature, a user invoked function would be able to access the module name from the state class. This would be possible.

```
def reset_data(state):
    if "page1" in state._module:
        state.df = df
        state.filter_value = 0
        notify(state, message="Table data is reset")
    if "page2" in state._module:
        state.chart_data = df
        state.fv = 0
        notify(state, message="Chart data is reset")
```

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #316, opened 2022-06-28
access module context from state class (#315)

dinhlongviolin1:
Solve Issue #315 

This would allow a function to be declared on the main module and that function can be used accross multiple modules
```
def reset_data(state):
    if "page1" in state._module:
        state.df = df
        state.filter_value = 0
        notify(state, message="Table data is reset")
    if "page2" in state._module:
        state.chart_data = df
        state.fv = 0
        notify(state, message="Chart data is reset")
```

### DIFF
--- a/taipy/gui/gui_actions.py
+++ b/taipy/gui/gui_actions.py
@@ -136,6 +136,26 @@ def get_context_id(state: State) -> t.Optional[str]:
 
 
 def get_module_name_from_state(state: State) -> t.Optional[str]:
+    """Get the module name that triggered a callback.
+
+    Pages can be defined in different modules yet share callbacks declared elsewhere (typically, the
+    application's main module).
+
+    This function returns the name of the module where the page (that holds the control that triggered the
+    callback) was declared.  This lets applications implement different behaviors depending on what page
+    is involved.
+
+    This function must be called only in the body of a callback function.
+
+    Arguments:
+
+        state: (State^): The `State` instance, which is an argument of the callback function.
+
+    Returns:
+
+        str: The name of the module that holds the definition of the page containing the control that
+            triggered this callback.
+    """
     if state and isinstance(state._gui, Gui):
         return state._gui._get_locals_context()
     return None
--- a/taipy/gui/gui_actions.py
+++ b/taipy/gui/gui_actions.py
@@ -135,7 +135,7 @@ def get_context_id(state: State) -> t.Optional[str]:
     return None
 
 
-def get_page_scope(state: State) -> t.Optional[str]:
+def get_module_name_from_state(state: State) -> t.Optional[str]:
     if state and isinstance(state._gui, Gui):
         return state._gui._get_locals_context()
     return None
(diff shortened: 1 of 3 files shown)


taipy/gui/gui_actions.py
```diff
@@ -134,7 +134,16 @@ def get_context_id(state: State) -> t.Optional[str]:
         return state._gui._get_client_id()
     return None
 
-def invoke_state_callback(gui: Gui, context_id: str, user_callback: t.Callable, args: t.Union[t.Tuple, t.List]) -> t.Any:
+
+def get_page_scope(state: State) -> t.Optional[str]:
+    if state and isinstance(state._gui, Gui):
+        return state._gui._get_locals_context()
```
FredLL-Avaiga, line 160: is this a str ?

FredLL-Avaiga (commented):
I'm still not sure I get it ...
gui._get_locals_context() returns the "module" for the UI modified variable ...
is there a module for on_action ?
What happens when the change comes from the back ?

taipy/gui/gui_actions.py
dinhlongviolin1: Yes it is a string

FredLL-Avaiga (approved):

taipy/gui/gui_actions.py
FabienLelaquais: Doc:
```
"""Get the module name that trigged a callback.

Pages can be defined in different modules yet share callbacks declared elsewhere (typically, the
application's main module).

This function returns the name of the module where the page (that holds the control that triggered
the callback) was declared.  This lets applications implement different behaviors depending on what page
is involved.

This function must be called only in the body of a callback function.

Arguments:

    state: (State^): The `State` instance, which is an argument of the callback function.

Returns:

    str: The name of the module that holds the definition of the page containing the control that triggered
          this callback.
"""
```

FabienLelaquais (approved):
👍

taipy/gui/gui_actions.py
```diff
@@ -134,7 +134,16 @@ def get_context_id(state: State) -> t.Optional[str]:
         return state._gui._get_client_id()
     return None
 
-def invoke_state_callback(gui: Gui, context_id: str, user_callback: t.Callable, args: t.Union[t.Tuple, t.List]) -> t.Any:
+
+def get_module_name_from_state(state: State) -> t.Optional[str]:
```
FabienLelaquais, line 138: Doc:
```
"""Get the module name that trigged a callback.

    Pages can be defined in different modules yet share callbacks declared elsewhere (typically, the
    application's main module).

    This function returns the name of the module where the page (that holds the control that triggered the
    callback) was declared.  This lets applications implement different behaviors depending on what page
    is involved.

    This function must be called only in the body of a callback function.

    Arguments:

        state: (State^): The `State` instance, which is an argument of the callback function.

    Returns:

        str: The name of the module that holds the definition of the page containing the control that
            triggered this callback.
"""
```

Issue #321, opened 2022-07-07
Implement rdp algorithm to chart data filtering


dinhlongviolin1:
**Description**
The RDP algorithm has already been in the product but has not been integrated into chart filtering. Now is the right time to do so!

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #322, opened 2022-07-07
Integrate RDP algorithm into chart filtering (#321)

dinhlongviolin1:
Solve Issue #321

A simple demo to test things out

```
import pandas as pd
import numpy as np
from taipy.gui import Gui

df = pd.DataFrame(np.random.randint(0, 10000000, size=(3000000, 1)), columns=["A"])
df['index1'] = df.index

chart_data = df.head(10)

def get_all_data(state):
    state.chart_data = df

def reset_data(state):
    state.chart_data = df.head(0)

Gui(page="""
<|Get All Data|button|on_action=get_all_data|>

<|Reset|button|on_action=reset_data|>

<|{chart_data}|chart|x=index1|y=A|limit_rows|limit_threshold=300|>
""").run(run_browser=False)

```


gui/src/components/Taipy/Chart.tsx
```diff
@@ -122,6 +123,7 @@ const Chart = (props: ChartProp) => {
         tp_onRangeChange,
         propagate = true,
         limitRows = false,
+        limitThreshold = undefined,
```
FredLL-Avaiga: `= undefined` is not necessary

taipy/gui/data/pandas_data_accessor.py
```diff
@@ -238,10 +239,12 @@ def __get_data(  # noqa: C901
             nb_rows_max = payload.get("width")
             # view with the requested columns
             if nb_rows_max and nb_rows_max < len(value) / 2:
-                value = value.copy()
-                value["tp_index"] = value.index
-                # we need to be more clever than this :-)
-                value = value.iloc[:: (len(value) // nb_rows_max)]
+                x_column, y_column = columns[1] if len(columns) > 1 else None, columns[0]
+                try:
+                    threshold = payload.get("limitThreshold", 50)
```
FredLL-Avaiga: the 50 value could be defined in the config (give a global default value)

FredLL-Avaiga (approved):
looks good to me

gui/src/components/Taipy/Chart.tsx
dinhlongviolin1: Thanks!

taipy/gui/data/pandas_data_accessor.py
dinhlongviolin1: You are absolutely right!

FabienLelaquais (approved):
👌

Issue #27, opened 2021-12-20
Allow filtering on table columns


FabienLelaquais:
**What would that feature address**

We want to allow users to select a filter from a table header to focus on a subset of the data.
The filter itself will pick one or several existing value and discard the rows that don't match.

***Description of the ideal solution***

Excel-like filtering.

***Caveats***

Some complex filters, like on numerical fields, might have to be implemented on the server.
Then the problem is how to select those filters from the front-end.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Create a demo that shows the feature
- [ ] Ensure documentaion is updated

FredLL-Avaiga:
Not sure client filtering would bring any value.
I think we should implement a back-end filter (inthe same way as the sort is done)
For the UI I propose to show a filter icon if filtering is available (property filter[col nome]=True or filter=True)
The icon would be different if a filter is set 
a small popover would open on click and allow the filter to be entered
filter should be different depending on the type of the column 
- text => filter by start if simple text and support for wildcards
- numeric or date => less than,  greater than 
- support for not and or

Another possibility would be to have a global popover for managing filter for the whole table with the possibility to add different filters on different columns combined with and/or ...

Any thought @FabienLelaquais ?

FabienLelaquais:
In my opinion, Client filtering does bring value. For broad data discovery.
Now I like your approach, which addresses this use case.
The 'or' or 'not' operator seem a bit overkill though.

And I also think that back-end filtering would be... code. Resulting on another DataFrame. I suspect we can live a long time without it for the time being.

FredLL-Avaiga:
> In my opinion, Client filtering does bring value. For broad data discovery.
> Now I like your approach, which addresses this use case.
> The 'or' or 'not' operator seem a bit overkill though.
> 
> And I also think that back-end filtering would be... code. Resulting on another DataFrame. I suspect we can live a long time without it for the time being.
> 

Front end filtering is complicated with pagination...

PR #324, opened 2022-07-07
table filters (#27)

FredLL-Avaiga:
filter or filter[col name] boolean property in table
So you can activate filtering on every column less one if needed or a specific subset of columns

Try it and let me know what you think

### DIFF
--- a/gui/src/components/Taipy/PaginatedTable.tsx
+++ b/gui/src/components/Taipy/PaginatedTable.tsx
@@ -53,7 +53,7 @@ import {
     ColumnDesc,
 } from "./tableUtils";
 import { useDispatchRequestUpdateOnFirstRender, useDynamicProperty, useFormatConfig } from "../../utils/hooks";
-import { FilterDesc, TableFilter } from "./TableFilter";
+import TableFilter, { FilterDesc } from "./TableFilter";
 
 const loadingStyle: CSSProperties = { width: "100%", height: "3em", textAlign: "right", verticalAlign: "center" };
 const skelSx = { width: "100%", height: "3em" };
--- a/gui/src/components/Taipy/PaginatedTable.tsx
+++ b/gui/src/components/Taipy/PaginatedTable.tsx
@@ -84,6 +84,7 @@ const PaginatedTable = (props: TaipyPaginatedTableProps) => {
     const [orderBy, setOrderBy] = useState("");
     const [loading, setLoading] = useState(true);
     const [aggregates, setAggregates] = useState<string[]>([]);
+    const [appliedFilters, setAppliedFilters] = useState<FilterDesc[]>([]);
     const { dispatch } = useContext(TaipyContext);
     const pageKey = useRef("no-page");
     const selectedRowRef = useRef<HTMLTableRowElement|null>(null);
@@ -158,7 +159,7 @@ const PaginatedTable = (props: TaipyPaginatedTableProps) => {
               }, "-agg")
             : "";
         const cols = colsOrder.map((col) => columns[col].dfid);
-        pageKey.current = `${startIndex}-${endIndex}-${cols.join()}-${orderBy}-${order}${agg}`;
+        pageKey.current = `${startIndex}-${endIndex}-${cols.join()}-${orderBy}-${order}${agg}${appliedFilters.map(af => `${af.col}${af.action}${af.value}`)}`;
         if (refresh || !props.data || props.data[pageKey.current] === undefined) {
             setLoading(true);
             const applies = aggregates.length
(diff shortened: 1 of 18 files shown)


gui/src/components/Taipy/PaginatedTable.tsx
```diff
@@ -318,14 +333,25 @@ const PaginatedTable = (props: TaipyPaginatedTableProps) => {
...
                                         >
                                             {columns[col].dfid === EDIT_COL ? (
-                                                active && editable && tp_onAdd ? (
-                                                    <IconButton onClick={onAddRowClick} size="small" sx={iconInRowSx}>
-                                                        <AddIcon />
-                                                    </IconButton>
-                                                ) : null
+                                                [
+                                                    active && editable && tp_onAdd ? (
+                                                        <Tooltip title="Add a row" key="addARow">
```
FabienLelaquais, line 341: Is there any smart mechanism that would make it possible to retrieve these text strings from somewhere else?
Thinking localization, obviously...

gui/src/components/Taipy/TableFilter.tsx
```diff
@@ -0,0 +1,354 @@
...
+    setFilter: (idx: number, fd: FilterDesc, remove?: boolean) => void;
+}
+
+const anchorOrigin = {
+    vertical: "bottom",
+    horizontal: "right",
+} as PopoverOrigin;
+
+const actionsByType = {
+    string: { "==": "equal", contains: "contains", "!=": "not equal" },
```
FabienLelaquais: "equal" -> "equals"?
"not equal" -> "is not", "does not equal"?

gui/src/components/Taipy/TableFilter.tsx
```diff
@@ -0,0 +1,354 @@
...
+    number: {
```
FabienLelaquais, line 51: Same here.
May be a short chat with Martin?

gui/src/components/Taipy/TableFilter.tsx
```diff
@@ -0,0 +1,354 @@
...
+                            setFilter={updateFilter}
+                        />
+                    ))}
+                    <FilterRow
+                        idx={-(filters.length + 1)}
+                        columns={props.columns}
+                        colsOrder={props.colsOrder}
+                        setFilter={updateFilter}
+                    />
+                    <Box sx={buttonBoxSx}>
```
FabienLelaquais, line 350: Should these button appear only when relevant? Like when there are no filters yet, what happens?

dinhlongviolin1 (approved):
works like a charm

gui/src/components/Taipy/TableFilter.tsx
```diff
@@ -0,0 +1,364 @@
...
+                        <Button endIcon={<ClearIcon />} onClick={onRemove}>{`Reset list (remove applied filter${
```
FabienLelaquais: I couldn't spot the case where there are no filters yet.

gui/src/components/Taipy/TableFilter.tsx
```diff
@@ -0,0 +1,364 @@
...
+                            filters.length > 1 ? "s" : ""
+                        }`}</Button>
+                    </Box>
+                </Box>
+            </Popover>
+        </>
+    );
+};
+
+export default TableFilter;
```
FabienLelaquais: Missing LF?

FabienLelaquais (approved):
Amazing!

gui/src/components/Taipy/PaginatedTable.tsx
FredLL-Avaiga: yes, we could use a localization library 
I guess we would need a global solution :-)

gui/src/components/Taipy/TableFilter.tsx
FredLL-Avaiga: I'll contact him

gui/src/components/Taipy/TableFilter.tsx
FredLL-Avaiga: the aim of this button is to remove all the applied filters: same as deleting all filters and then clicking apply but without removing the filters from the list

gui/src/components/Taipy/TableFilter.tsx
FredLL-Avaiga: mypy didn't complain but ok

gui/src/components/Taipy/TableFilter.tsx
FredLL-Avaiga: don't really like disappearing buttons ...

FabienLelaquais: Inactive?

gui/src/components/Taipy/TableFilter.tsx
FabienLelaquais: I'll check this out when it runs on my screen...

gui/src/components/Taipy/TableFilter.tsx
FredLL-Avaiga: OK

Issue #323, opened 2022-07-07
BUG-Migrating an application from 1.0 to 1.1 may produce Axios error messages


FabienLelaquais:
**Description**

The client identifier that Taipy creates to associate a server and a client is stored in the localStorage of the browser.
When you run the application, a Permanent Redirect request is created, based on that Client Id.
Because this is stored in the browser cache, and the redirection is not similar in 1.0 compared to 1.1, the app fails to render the page properly.

**How to reproduce**

- Create a single-page application and run it against Taipy GUI 1.0.
- Migrate the Taipy GUI package to 1.1.
- re-run the app.
Accessing the page generates an Axios error message instead of the page content.

- Back to Taipy GUI 1.0, and everything works fine.
- Clear the browser cache and the app runs fine as well.

**Expected behavior**

Upgrading Taipy GUI should not have any bad impact.

PR #325, opened 2022-07-07
#323 add a version parameter to the request

FredLL-Avaiga:
so that we won't get permanent redirect kicking in from 1.0
This should hold the real (read from back-end) value in sunbsequent versions

### DIFF
--- a/gui/src/components/pages/TaipyRendered.tsx
+++ b/gui/src/components/pages/TaipyRendered.tsx
@@ -61,7 +61,7 @@ const TaipyRendered = (props: TaipyRenderedProps) => {
             dispatch(createPartialAction(path.slice(1), false));
         } else {
             axios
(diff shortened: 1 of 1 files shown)


dinhlongviolin1 (approved):

gui/src/components/pages/TaipyRendered.tsx
```diff
@@ -61,7 +61,7 @@ const TaipyRendered = (props: TaipyRenderedProps) => {
             dispatch(createPartialAction(path.slice(1), false));
         } else {
             axios
-            .get<AxiosRenderer>(`/taipy-jsx${path}`, {params: {client_id: state.id || ""}})
+            .get<AxiosRenderer>(`/taipy-jsx${path}`, {params: {client_id: state.id || "", v: "1.1"}})
```
FabienLelaquais, line 64: Wouldn't it be clearer if client_id was simply dropped from the params?

FabienLelaquais (approved):
👌

Issue #329, opened 2022-07-15
Visual Element Extension back end


FredLL-Avaiga:
**Description**
Design and implement a first version of Visual Element Extension for Backend

Ie a way for the developper to add a component with attributes

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #334, opened 2022-07-15
#329 extension backend

FredLL-Avaiga:
First step in extension: To be discussed

How to use:

```
from pathlib import Path
from taipy.gui import Gui
from taipy.gui.extension import ElementLibrary, Element, ElementAttribute, PropertyType

import typing as t

val = None

md = """
# Custom test
<|{val}|mylib.fred|multiline|>

<|{val}|input|>

<|change val|button|on_action={lambda s: s.assign("val", "val other")}|>
"""

class MyLibrary(ElementLibrary):

    elts = [Element("fred", "value", [ElementAttribute("value", PropertyType.dynamic_string, "Fred"), ElementAttribute("multiline", PropertyType.boolean, False)], "Input")]

    def get_name(self) -> str:
        return "mylib"

    def get_elements(self) -> t.List[Element]:
        return MyLibrary.elts

    def get_scripts(self) -> t.List[str]:
        return ["mylib1.js", "mylib2.js"]

    def get_styles(self) -> t.List[str]:
        return None

    def get_resource(self, name: str) -> Path:
        return Path(".") / f"demo-custom.{name.split('.')[-1]}"

Gui.add_library(MyLibrary())
Gui(md).run(use_reloader=False, async_mode="threading")
```


taipy/gui/renderers/factory.py
```diff
@@ -513,6 +519,27 @@ class _Factory:
...
+            if builder is not None:
+                prop = builder.default_attribute
+        return prop
+
+    @staticmethod
+    def call_builder(gui: "Gui", name: str, all_properties: t.Optional[t.Dict[str, t.Any]] = None) -> t.Optional[_Builder]:
+        builder = _Factory.__CONTROL_BUILDERS.get(name)
+        if builder is None:
+            control = _Factory.__USER_CONTROLS.get(name)
+            if control is not None:
```
dinhlongviolin1: Verify control type or check whether call_builder method is available

taipy/gui/renderers/user_control.py
```diff
@@ -0,0 +1,88 @@
...
+
+from ..types import _AttributeType
+from .utils import _to_camel_case
+from .builder import _Builder
+
+if t.TYPE_CHECKING:
+    from ..gui import Gui
+
+
+class UserAttribute(ABC):
```
dinhlongviolin1: Not an ABC class since it will be used directly by the user

taipy/gui/renderers/user_control.py
```diff
@@ -0,0 +1,88 @@
...
+        if not isinstance(self.name, str) or not self.name or not self.name.isidentifier():
+            warnings.warn(f"User Control '{control}' should have a valid attribute name '{self.name}'")
+        if not isinstance(self.attribute_type, _AttributeType):
+            warnings.warn(f"User Attribute '{control}.{self.name}' should have a valid type '{self.attribute_type}'")
+    
+    def get_tuple(self) -> tuple:
+        return (self.name, self.attribute_type, self.default_value)
+
+
+class UserControl(ABC):
```
dinhlongviolin1: idem

dinhlongviolin1 (approved):
Good first step, just a few comments

taipy/gui/gui.py
```diff
@@ -230,6 +232,9 @@ def __init__(
         if env_filename is not None:
             self.__env_filename = env_filename
 
+    def add_components(self, user_controls: t.List[UserControl]):
```
dinhlongviolin1: Maybe verify usercontrol here

taipy/gui/renderers/factory.py
```diff
@@ -513,6 +519,27 @@ class _Factory:
     # TODO: process \" in property value
     _PROPERTY_RE = re.compile(r"\s+([a-zA-Z][\.a-zA-Z_$0-9]*(?:\[(?:.*?)\])?)=\"((?:(?:(?<=\\)\")|[^\"])*)\"")
 
+    @staticmethod
+    def set_user_builders(user_builders: t.Dict[str, UserControl]):
```
dinhlongviolin1: or verify UserControl here

taipy/gui/renderers/user_control.py
FredLL-Avaiga: true that

FabienLelaquais (changes requested):
That's a great step!

I however recommend that the user's additional visual elements are stored in dedicated libraries.
I will refer to this notion as an 'Element Library' or 'Visual Element Library' in the rest of this document.

Element Libraries allow creating applications that use several libraries, coming from different places.
It also allows for grouping elements by provider (I'm thinking google.maps, google.earth...).

Element Libraries hold a list of user-defined elements and probably can have what it takes to locate the JavaScript bundle that defines the actual components.

- An element library has a user-defined name ("taipy" is reserved) that must be a valid Python identifier.
  This name is used as the element name in the page definition: <library_name>.<component_name> in Markdown, and <library_name>:<component_name> in HTML.
- Each element in a library has a user-defined name  that must be a valid Python identifier.
	- Each element holds a list of properties: an element property has a name (a valid Python identifier), a type and a default value.
	- Each element declares a 'default property name' that indicates where to store the first fragment of the element Markdown syntax or the text value of the HTML syntax.
	- Each element must define a function that generates the React component instantiation (essentially the mapping between the element's properties and the React component's).

Opened questions:

- Callbacks: can users define element-specific parameters (like the table-specific callbacks)
- JavaScript:  How to bundle and distribute a Visual Elements Library?

taipy/gui/gui.py
```diff
@@ -230,6 +232,13 @@ def __init__(
         if env_filename is not None:
             self.__env_filename = env_filename
 
+    @staticmethod
+    def add_components(user_controls: t.List[UserControl]):
```
FabienLelaquais: That would be add_element or/and add_element_library or something

taipy/gui/renderers/user_control.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2022 Avaiga Private Limited
```
FabienLelaquais: I suspect we need a specific package for these.
taipy.gui.extension?

Then taipy.gui.extension.Element and ElementLibrary

FabienLelaquais (changes requested):
See comments

FabienLelaquais (approved):

FabienLelaquais:
I really like the Library class. Makes sense.
I would not have stored the element definitions in an array though. No need to iterate, dict sounds like a fit.

And the Attribute* classes will have to be exposed and renamed to Property*

FredLL-Avaiga:
> I really like the Library class. Makes sense. I would not have stored the element definitions in an array though. No need to iterate, dict sounds like a fit.
> 
> And the Attribute* classes will have to be exposed and renamed to Property*

Element and ElementAttribute classes are a way to force the structure of the data to provide in a clean and explicit way I think

PR #341, opened 2022-07-25
gui config (#326)

dinhlongviolin1:
Solve Issue #326 
- Move taipy into src folder
- fix tests by modyfing sys.modules
- fix additonal codeql path vulnerabilities


setup.py
```diff
@@ -37,6 +37,7 @@
     "flask-talisman>=1.0,<2.0",
     "gevent>=21.12.0,<22.0",
     "gevent-websocket>=0.10.1,<0.11",
+    "taipy-config@git+https://git@github.com/Avaiga/taipy-config.git@develop",
```
FredLL-Avaiga, line 40: ?

FredLL-Avaiga (approved):
get taipy-config from git@develop ?

setup.py
dinhlongviolin1: You are right. We are getting taipy-config from github (same for taipy-core). We will change this as soon as taipy-config is pushed to pypi

FabienLelaquais (approved):
Looking really good.
Let me update the doc build process to reflect that...

Issue #320, opened 2022-07-06
BUG-Notebook-Need-gui.stop()


vhfgosselin:
**Description**
Under different notebook environments (I tried under jupyter and also watson studio), if I execute a cell containing taipy code, it works but if I decide to rerun the same cell I do get the following error message:
ConnectionError                           Traceback (most recent call last)
Input In [3], in <cell line: 15>()
     12 value = 10
     14 gui = Gui(page)
---> 15 gui.run(ngrok_token="2AfMLycYRW891ajNoVjRvb4muex_6HFojpe17gxbd3Ewgn959")

File ~\anaconda3\envs\new_taipy_11\lib\site-packages\taipy\gui\gui.py:1409, in Gui.run(self, run_server, run_in_thread, ssl_context, **kwargs)
   1406 if not run_server:
   1407     return self.get_flask_app()
-> 1409 return self._server.runWithWS(
   1410     host=app_config["host"],
   1411     port=app_config["port"],
   1412     debug=app_config["debug"],
   1413     use_reloader=app_config["use_reloader"],
   1414     flask_log=app_config["flask_log"],
   1415     run_in_thread=run_in_thread,
   1416     ssl_context=ssl_context,
   1417 )

File ~\anaconda3\envs\new_taipy_11\lib\site-packages\taipy\gui\server.py:178, in _Server.runWithWS(self, host, port, debug, use_reloader, flask_log, run_in_thread, ssl_context)
    176     sock.close()
    177     if result == 0:
--> 178         raise ConnectionError(
    179             f"Port {port} is already opened on {host_value}. You have another server application running on the same port."
    180         )
    181 if not flask_log:
    182     log = logging.getLogger("werkzeug")

ConnectionError: Port 5000 is already opened on 127.0.0.1. You have another server application running on the same port.

**How to reproduce**
A code fragment
```
from taipy.gui import Gui, Markdown
from pyngrok import ngrok

page = Markdown("""
# Taipy in a Notebook

Value: <|{value}|>

Set: <|{value}|slider|>
""")
value = 10
gui = Gui(page)
gui.run(ngrok_token="2AfMLycYRW891ajNoVjRvb4muex_6HFojpe17gxbd3Ewgn959")
...
```

**Expected behavior**
As would have expected no error. The problem doesn't occur if I do execute in the following cell gui.stop()
However, in the documentation, it is written that the gui.stop() is only required when for example an on_change() is created after gui.run() in the subsequent cell. 
[...]

PR #343, opened 2022-07-29
Fix Gui Stop in Notebook (#320)

dinhlongviolin1:
Solve Bug #320


FabienLelaquais:
I see where you're going... but I don't think we're done, are we?

Storing the Gui-port association is a starting point, but then we need to use it when 'run()' is invoked.
We also probably want to remove the association when the Gui is stopped...
Am I missing something??
Thanks!

FredLL-Avaiga (approved):

dinhlongviolin1:
> I see where you're going... but I don't think we're done, are we?
> 
> Storing the Gui-port association is a starting point, but then we need to use it when 'run()' is invoked. We also probably want to remove the association when the Gui is stopped... Am I missing something?? Thanks!

You are right, it should be removed when the Gui is stopped. But in our use case, we don't have to do it since it will only be checked/executed on run() method.

FabienLelaquais (approved):
👍

PR #344, opened 2022-07-29
Minor doc changes and fixes.

FabienLelaquais:


FredLL-Avaiga (approved):

src/taipy/gui/gui.py
```diff
@@ -1301,19 +1301,31 @@ def run(
                 Note that if you are running in an IPython notebook context, the Web
                 server is always run in a separate thread.
             ssl_context (Optional[Union[ssl.SSLContext, Tuple[str, Optional[str]], te.Literal['adhoc']]]):
```
dinhlongviolin1: we are not using ssl_context at the moment. Fred removed it in one of the PR if I remember correctly

dinhlongviolin1 (approved):
Looks all good, just a comment on ssl

src/taipy/gui/gui.py
FabienLelaquais: Dropped.

Issue #332, opened 2022-07-15
Add taipy-gui version to requests


FredLL-Avaiga:
**What would that feature address**
That would allow easier version migration

***Description of the ideal solution***
Manage a global version that would be available in backend and frontend

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
version.json needs to be in the taipy directory so that it can be found at runtime
`...python3.9/site-packages/taipy/gui/gui.py:221: UserWarning: Cannot retrieve version.json file: [Errno 2] No such file or directory: '...python3.9/site-packages/version.json'`

FredLL-Avaiga:
put it in taipy/gui so that there's no risk of file override from different taipy packages

PR #346, opened 2022-08-01
on navigate (#332)

FredLL-Avaiga:
```
from src.taipy.gui import Gui, State, Markdown

md_1 = """
This is page 1
<|{count}|number|label=counter|>
"""

md_2 = """
This is page 2
"""

md_3 = """
This is page 3
"""

md = """
# on_navigate test
<|navbar|lov={navlov}|>

<|content|>
"""

navlov = [
    ("/page1", "Page 1"),
    ("/page2", "Page 2"),
]

pages = {
    "/": Markdown(md),
    "page1": Markdown(md_1),
    "page2": Markdown(md_2),
    "page3": Markdown(md_3),
}

count = 0

def on_navigate(state: State, page: str) -> str:
    if page == "page2" and state.count == 1:
        return "page3"
    elif state.count == 2:
        return None
    return page

Gui(pages=pages).run(use_reloader=False)
```

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -96,6 +96,9 @@ class Gui:
         on_init (Callable): The function that is called on the first connection of a new user.<br/>
             It defaults to the `on_init()` global function defined in the Python
             application.
+        on_navigate (Callable): The function that is called when a page is requested.<br/>
+            It defaults to the `on_navigate()` global function defined in the Python
+            application.
         state (State^): **Only defined when running in an IPython notebook context.**<br/>
             The unique instance of `State^` that you can use to change bound variables
             directly, potentially impacting the interface in real-time.
@@ -204,6 +207,7 @@ class Gui:
         self.on_action: t.Optional[t.Callable] = None
         self.on_change: t.Optional[t.Callable] = None
         self.on_init: t.Optional[t.Callable] = None
+        self.on_navigate: t.Optional[t.Callable] = None
 
         # sid from client_id
         self.__client_id_2_sid: t.Dict[str, t.Set[str]] = {}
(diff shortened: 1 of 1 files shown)


src/taipy/gui/gui.py
```diff
@@ -96,6 +96,9 @@ class Gui:
         on_init (Callable): The function that is called on the first connection of a new user.<br/>
             It defaults to the `on_init()` global function defined in the Python
             application.
+        on_navigate (Callable): The function that is called when a page is requested.<br/>
```
FabienLelaquais, line 100: "It defaults to the `on_navigate()` global function defined in the Python  application."

I would add, just like for *on_init()*, that if there is no such function, nothing happens.

src/taipy/gui/gui.py
```diff
@@ -1186,22 +1190,31 @@ def __init_route(self):
...
+        nav_page = page_name
+        if hasattr(self, "on_navigate") and callable(self.on_navigate):
+            try:
+                nav_page = self.on_navigate(self.__get_state(), page_name)
+                if not isinstance(nav_page, str):
+                    warnings.warn(f"on_navigate returned a invalid page name '{nav_page}'")
+                    nav_page = page_name
+            except Exception as e:
+                warnings.warn(f"Exception on on_navigate execution\n{e}")
+        page = next((page_i for page_i in self._config.pages if page_i._route == nav_page), None)
```
FabienLelaquais, line 1209: If nav_page != page_name, you should I think call 'on_navigate' again...

FabienLelaquais (changes requested):
Small yet important change...

src/taipy/gui/gui.py
FredLL-Avaiga: Why would I do this ?
if new_page != page the dev changed it in on_navigate ... If I call again on_navigate, how can the dev differentiate from a "normal" call ?

FabienLelaquais: In my mind, it was to simplify the user's code, if there are several tests to perform before choosing a page name or another.
When I have a real use case, I'll let you know. Forget it for the time being.
Thx

FabienLelaquais (approved):
👍

dinhlongviolin1 (approved):
LGTM!

Issue #352, opened 2022-08-04
Improve decimation algorithm for taipy-gui and make decimation class extensible


dinhlongviolin1:
**Description**
As of today, only RDP algorithm is implemented. Other decimation algorithm might be more effecient than RDP in certain use cases. 

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #353, opened 2022-08-04
Gui Decimator (#352)

dinhlongviolin1:
Solve Issue #352 

```
import pandas as pd
import numpy as np
from src.taipy.gui import Gui
from src.taipy.gui.data.decimator import MinMaxDecimator, RDP, LTTB

df = pd.DataFrame(np.random.randint(0, 10000000, size=(30000, 1)), columns=["A"])
df["index1"] = df.index

# decimator_class = RDP(epsilon=100)
# decimator_class = RDP(n_out=300)
# decimator_class = LTTB(n_out=300)
# decimator_class = MinMaxDecimator(n_out=300)

gui = Gui(
    page=""" 
<|{df}|chart|x=index1|y=A|limit_rows|decimator=decimator_class|>
"""
)
gui.run()

```


src/taipy/gui/gui.py
```diff
@@ -734,6 +734,16 @@ def _get_user_function(self, func_name: str) -> t.Union[t.Callable, str]:
             return func
         return func_name
 
+    def _get_user_class(self, class_name: str, class_type: type) -> t.Union[object, str]:
```
FredLL-Avaiga: _get_user_instance ?

src/taipy/gui/data/pandas_data_accessor.py
```diff
@@ -264,14 +265,19 @@ def __get_data(  # noqa: C901
             )
         else:
             ret_payload["alldata"] = True
+            decimator = payload.get("decimator", None)
+            decimator_class = (
```
FredLL-Avaiga: decimator_instance ?

FredLL-Avaiga (approved):
Should work :-)

src/taipy/gui/gui.py
dinhlongviolin1: make sense! thx

src/taipy/gui/data/pandas_data_accessor.py
dinhlongviolin1: agree!

src/taipy/gui/data/utils.py
```diff
@@ -12,16 +12,43 @@
...
+    def _is_applicable(self, data: t.Any, nb_rows_max: int):
+        if self.applied_threshold is None:
+            if nb_rows_max < len(data):
+                return True
+        elif self.applied_threshold < len(data):
+            return True
+        return False
+
+    @abstractmethod
+    def decimate(self, data: np.ndarray) -> np.ndarray:
```
FabienLelaquais, line 38: We'll need a bit more explanation at this point.
That IS the real magic of all of this.

src/taipy/gui/renderers/factory.py
```diff
@@ -97,7 +97,7 @@ class _Factory:
                 ("on_range_change", PropertyType.function),
                 ("active", PropertyType.dynamic_boolean, True),
                 ("limit_rows", PropertyType.boolean),
```
FabienLelaquais: Do we still need 'limit_rows`?

FabienLelaquais (approved):
That looks really, really neat!

src/taipy/gui/renderers/factory.py
dinhlongviolin1: nice catch! no we do not

src/taipy/gui/data/utils.py
dinhlongviolin1: Can you suggest anything to add

FabienLelaquais: First of all, I would rename the class 'Decimator', for short. The ABC aspect is an implementation detail.
And this class should definitively get some doc. At least a skeleton, full of TODO notes.

Then, this method really is the entry point of that feature. So we will need to document it thoughtfully. 
Again, TODO entries is fine to me.

Issue #330, opened 2022-07-15
Visual Element Extension front end


FredLL-Avaiga:
**What would that feature address**
Allow the developper to add a visual element as a js library

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #354, opened 2022-08-05
#330 adding a UI component in the front WIP

FredLL-Avaiga:
A big re-organization of the front-end js libraries so that a dev can add components
- taipy-gui js is now cut in 3 parts:
  - TaipyGui: holds the main taipy-gui sources and export what's needed for a 3rd party library
  - TaipyVendor holds the commonly used library (that should not be duplicated) ie react, mui 
  - taipy-gui-dom.js uses the above 2 libraries and instantiate the main component so that the page can be displayed

A third party library developper should then: 
- define the components on the backend : library name, component names, js function name to register the components
- create a library of js components that uses TaipyGui and TaipyVendor libraries which provides
  - a register function
  - the react components

PS : I have an example of a 3rd party library. I suppose it should be provided as an example/starter ...


FabienLelaquais (approved):
Very, very promising. Great job!

dinhlongviolin1 (approved):
I must admit I dont fully understand everything but only partly. But it works like a charm so I am gonna leave my trust to you. Amazing work!

Issue #356, opened 2022-08-09
BUG-dict vs page scope


FredLL-Avaiga:
**Description**
dict access don't work right with page scopes

**How to reproduce**

```
from src.taipy.gui import Gui, State

a_key = "key"
a_dict = {a_key: "key in dict"}

def on_action(state: State):
    print(state.a_key)
    state.a_dict.key = {"l": len(state.a_dict)}
    print(state.a_dict.keys())
    state.a_dict[state.a_key] = {"l": len(state.a_dict)}
    print(state.a_dict.keys())

Gui("""
# test issue dict and keys vs state
<|{a_dict[a_key]}|input|>
<|{a_dict.key}|input|>
<|test|button|on_action=on_action|>
""").run()
```

**Expected behavior**
modifying via UI should modify state.a_dict.key
modifying via on_action should modify state.a_dict.key
No warning/error should be thrown

PR #358, opened 2022-08-12
fix gui map dict assignment (#356)

dinhlongviolin1:
Fix bug #356


src/taipy/gui/gui.py
```diff
@@ -437,6 +434,10 @@ def _update_var(
             self.__send_var_list_update(list(derived_modified), var_name)
 
     def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[str] = None):
+        suffix_var_name = ""
+        if "." in var_name:
+            suffix_var_name = var_name[var_name.index(".") + 1 :]
```
FredLL-Avaiga: use split ?
or at least cache the index result

FredLL-Avaiga (changes requested):
does this fix all the issues revealed by the test case ?
When I try :
```
<|{a_dict[a_key]}|input|>
<|{a_dict.key}|input|>
```
the 2 inputs don't show the same variable although a_key="key" :-(

src/taipy/gui/gui.py
```diff
@@ -408,10 +408,7 @@ def _update_var(
         propagate=True,
         holder: t.Optional[_TaipyBase] = None,
         on_change: t.Optional[str] = None,
-        from_map_dict: bool = False,
     ) -> None:
-        if from_map_dict:
-            var_name = _variable_encode(var_name, self._get_locals_context())
         if holder:
             var_name = holder.get_name()
         hash_expr = self.__evaluator.get_hash_from_expr(var_name)
```
FredLL-Avaiga, line 414: I guess the calculation of the right variable name taht is done in __call_on_change should be done somewhere here ;-)

src/taipy/gui/gui.py
dinhlongviolin1: thank you!

src/taipy/gui/gui.py
dinhlongviolin1: It is not neccessary anymore I think, I dont remember why I added that but that was when everything was a mess :D

src/taipy/gui/utils/_evaluator.py
```diff
@@ -249,7 +266,7 @@ def re_evaluate_expr(self, gui: Gui, var_name: str) -> t.Set[str]:
             expr_decoded, _ = _variable_decode(expr)
             hash_expr = self.__expr_to_hash.get(expr, "UnknownExpr")
             if expr != var_name and not expr.startswith("_Taipy"):
-                expr_var_map = self.__expr_to_var_map.get(expr)  # ["x", "y"]
+                expr_var_map = self.__expr_to_var_map.get(expr, {})  # ["x", "y"]
                 if expr_var_map is None:
```
FredLL-Avaiga: How can this be None ?

FredLL-Avaiga (approved):
Works for me

src/taipy/gui/utils/_evaluator.py
dinhlongviolin1: you are right, I will remove the empty dict assignment. I added that because mypy said expr_var_map has wrong type.

PR #360, opened 2022-08-16
replace killablethread with kthread

dinhlongviolin1:


FredLL-Avaiga (approved):
looks good

Pipfile
```diff
@@ -22,6 +22,8 @@ flask-talisman = "==1.0"
 gevent = "==21.12.0"
 gevent-websocket = "==0.10.1"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
+kthread = "==0.2.3"
+werkzeug = "==2.0.3"
```
FredLL-Avaiga, line 26: Why do we need to fix werkzeug version ?

dinhlongviolin1: in 2.1, they remove 'werkzeug.server.shutdown' so the thread couldnt shut down, which cause the sudden break in everything when stopping flask dev server. This does not affect the application when using gevent.

FabienLelaquais (approved):
👌

Issue #366, opened 2022-08-22
Table control -- Limit editing to user-specificed columns


mshell99:
**Description**
Provide users with the ability to limit editing to specific columns, syntax like  editable[col_names].
Often in tables some columns should never be edited by users (id numbers, names, etc.)  No way of limiting this now

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #365, opened 2022-08-22
BUG- Table Control -- Boolean data not shown when table is editable


mshell99:
**Description**
If a table control has a boolean column and the table is editable. The boolean "switch" indicator is not shown, only the edit pencil icon
**How to reproduce**
df is a DataFrame with a boolean column
md = """
<|{df}|table|on_edit=on_table_edit|>
"""

Dr-Irv:
I installed the current version of `taipy-gui` as of commit 45795dade1085dcb777fcdceee46ed72b81f0f44 which should include #367 .   Consider the following example:
```python
import pandas as pd
from taipy import Gui
from taipy.gui import State

df = pd.DataFrame({"x": [1, 2, 3], "flag": [True, False, True]})


def myedit(state: State, varname: str, action: str, payload: dict):
    temp_df = state.df.copy()
    temp_df.loc[payload["index"], payload["col"]] = payload["value"]
    state.df = temp_df


def printit(state: State):
    print(state.df)


page = """
<|{df}|table|height=max-content|width=max-content|show_all=True|on_edit=myedit|>

<|print|button|on_action=printit|>
"""


if __name__ == "__main__":
    Gui(page=page, css_file="first").run(async_mode="threading", dark_mode=False)
```
The table is editable, but for the boolean property, I can't use the switch to do the actual editing of the value.  If I do click on the pencil, I can change the value by typing it in, and the value is updated in the dataframe as illustrated by clicking on the "Print" button.  But the switch value is not updated  in the browser when the table value is updated.

@FredLL-Avaiga Can you look into this?

FredLL-Avaiga:
I suppose because the value typed and sent is a string.
A quick workaround would be to transform it to a boolean in your on_edit.
I'm not sure if we really want to cast the values before passing them to on_edit or simply pass everything as a string to allow more freedom ...

FredLL-Avaiga:
I suppose the right thing to do is provide both value (casted and original string)
I'm on it (and for numbers also)

FredLL-Avaiga:
The value is now casted to the type of the column and I've added a user_value: str (the value as entered by the user) to the payload

Issue #333, opened 2022-07-15
Add a multi line property for the input visual element


FlorianJacta:
**Description**
The goal would be to be able to write texts on multiple lines in the 'input' visual element. It can have multiple uses. Maybe, a new 'multiline' property has to be created for this purpose.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
3 new properties for input:
- multiline: boolean
- rows: optional integer
- cols: optional integer
@FabienLelaquais What do you think of the property names ?

FabienLelaquais:
I discussed, I really don't like those `rows`/`columns` names, although indeed these properties are the ones used in the `textarea` HTML tag.
What about `show_length` and `show_lines`? (`text` does not really work...)

FredLL-Avaiga:
width_in_rows / height_in_cols ?
Or width/height with some docs
https://mui.com/material-ui/api/text-field/
no cols in here but we have rows / minrows and maxrows

I really think we should expose multiline and rows (maybe use it as minRows) and be consistent with MUI and HTML @FabienLelaquais

FabienLelaquais:
My issue really is the language.
Line (for multiline), Row (for, well rows).

Our target is not HTML or MUI developers. Rather the contrary.
Rows is (for good reasons) heavily used in the context of tables...
And I can see nothing better than *multiline*.

FredLL-Avaiga:
line ? boolean ? meaning multi ?
row : optional integer ?

FabienLelaquais:
What I'm saying is that the same semantic object ("series of characters with no carriage return") is represented with two words.
- "Line" in "multiline"
- "Row" in "rows"

That's what I don't like.

I do understand the HTML+MUI argument. Just that I don't buy it.
I would certainly use 'line' everywhere.
-> "n_lines", "line_number", "lines_shown" (show_lines really sounds like a Boolean  property).

"Length" is usually what we would like for the x dimension. Makes no sense here.
"line_length", may be ?

FredLL-Avaiga:
I think the complexity here is that we want to define the "size" of the element and not limit the content ...
I think multiline is quite understandable
I suppose lines_shown is the most evident ...

FabienLelaquais:
Yeah yeah, I get that.

[...]

PR #367, opened 2022-08-22
Table and Input

FredLL-Avaiga:
- #366 specify editable by column
- #365 boolean editable column
- #333 input multiline


dinhlongviolin1 (approved):
Well done!

Issue #368, opened 2022-08-22
BUG- var_name in on_edit() for tables doesn't provide name of DataFrame


Dr-Irv:
**Description**
If you have an editor for a table, then the value of `var_name` passed into the `on_edit()` function doesn't identify the name of the DataFrame.

**How to reproduce**
A code fragment
```python
from taipy import Gui;
from taipy.gui.state import State
import pandas as pd
df = pd.DataFrame({"x":[1,2,3]})

def myedit(state: State, var_name: str, action: str, payload: dict):
    print(var_name)

page = "<|{df}|table|id=df|on_edit=myedit|>"

if __name__ == "__main__":
    Gui(page=page).run()
```

Then just edit an element and click the checkbox, and it will print `_TpD_tpec_TpExPr_df_TPMDL_2` .  

**Expected behavior**
I expected it would print `df`

**Screenshots**
N/A

**Runtime environment**
Please specify relevant indications.
 - OS: Windows 10
 - Browser: Chrome  104.0.5112.81

FredLL-Avaiga:
Hi @Dr-Irv, 
No need to add a tag property, you should get the right variable name in your on_edit function when the linked PR is merged

Dr-Irv:
> Hi @Dr-Irv, No need to add a tag property, you should get the right variable name in your on_edit function when the linked PR is merged

Great!

Dr-Irv:
I was able to pull your PR and install it and it worked.  Thanks.

PR #371, opened 2022-08-23
on_action user var name (#368)

FredLL-Avaiga:
retrieve original var name for on_action

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -28,8 +28,6 @@ import tzlocal
 from flask import Blueprint, Flask, g, jsonify, request, send_file, send_from_directory
 from werkzeug.utils import secure_filename
 
-from src.taipy.gui.utils.types import HOLDER_PREFIX_START, HOLDER_PREFIXES
-
 if util.find_spec("pyngrok"):
     from pyngrok import ngrok
 
@@ -72,6 +70,7 @@ from .utils import (
     _TaipyLovValue,
     _variable_decode,
 )
+from .utils.types import _HOLDER_PREFIX_START, _HOLDER_PREFIXES
 from .utils._adapter import _Adapter
 from .utils._bindings import _Bindings
 from .utils._evaluator import _Evaluator
@@ -435,8 +434,8 @@ class Gui:
 
     def __get_real_var_name(self, var_name: str) -> t.Tuple[str, str]:
         # Handle holder prefix if needed
-        if var_name.startswith(HOLDER_PREFIX_START):
-            for hp in HOLDER_PREFIXES:
+        if var_name.startswith(_HOLDER_PREFIX_START):
+            for hp in _HOLDER_PREFIXES:
                 if var_name.startswith(hp):
                     var_name = var_name[len(hp):]
                     break
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -433,6 +433,8 @@ class Gui:
             self.__send_var_list_update(list(derived_modified), var_name)
 
     def __get_real_var_name(self, var_name: str) -> t.Tuple[str, str]:
+        if not var_name:
+            return (var_name, var_name)
         # Handle holder prefix if needed
         if var_name.startswith(_HOLDER_PREFIX_START):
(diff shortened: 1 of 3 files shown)


src/taipy/gui/utils/types.py
```diff
@@ -103,28 +103,31 @@ def cast_value(self, value: t.Any):
...
+        return _HOLDER_PREFIX_START + "C"
 
 
 class _TaipyContentImage(_TaipyBase):
     @staticmethod
     def get_hash():
-        return "_TpCi"
+        return _HOLDER_PREFIX_START + "Ci"
+
+_HOLDER_PREFIX_START = "_Tp"
```
FabienLelaquais: Is there a matching HOLDER_PREFIX_END or something?
PREFIX really means... starting a string value of some sort.
I mean HOLDER_PREFIX should suffice.

src/taipy/gui/utils/types.py
```diff
@@ -45,13 +45,13 @@ def _get_readable_name(self):
 
     @staticmethod
     def get_hash():
-        return "TaipyBase"
+        return NotImplemented
```
FabienLelaquais, line 48: Shouldn't this be a raise instead of a return?

src/taipy/gui/gui.py
```diff
@@ -445,18 +452,24 @@ def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[st
             var_name = var_name_decode
         else:
             if var_name not in self.__var_dir._var_head:
-                warnings.warn(f"Can't find matching variable for {var_name} on {current_context} context")
-                return
+                raise NameError(f"Can't find matching variable for {var_name} on {current_context} context")
```
FabienLelaquais: I'm not sure but it seems like 'on context: {current_context}' might be easier to read

src/taipy/gui/gui.py
```diff
@@ -445,18 +452,24 @@ def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[st
...
             for k, v in self.__var_dir._var_head[var_name]:
                 if v == current_context:
                     var_name = k
                     _found = True
                     break
             if not _found:
-                warnings.warn(f"Can't find matching variable for {var_name} on {current_context} context")
-                return
-        var_name = f"{var_name}.{suffix_var_name}" if suffix_var_name else var_name
+                raise NameError(f"Can't find matching variable for {var_name} on {current_context} context")
```
FabienLelaquais: Same remark on error message

src/taipy/gui/gui.py
```diff
@@ -445,18 +452,24 @@ def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[st
...
+        return f"{var_name}.{suffix_var_name}" if suffix_var_name else var_name, current_context
+
+
+    def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[str] = None):
+        try:
+            var_name, current_context = self.__get_real_var_name(var_name)
+        except Exception as e:
+            warnings.warn(f"{e}")
```
FabienLelaquais, line 471: Is a warning enough at this point? May be an actual error?

FabienLelaquais (approved):
Minor changes, The actual issue seems to be properly handled. Well done!

src/taipy/gui/utils/types.py
FredLL-Avaiga: true

src/taipy/gui/gui.py
FredLL-Avaiga: ok

src/taipy/gui/gui.py
FredLL-Avaiga: I didn't touch this part (only refactor)

src/taipy/gui/utils/types.py
FredLL-Avaiga: no I think you can return NotImplemented or raise NotImplementedError
https://stackoverflow.com/questions/878943/why-return-notimplemented-instead-of-raising-notimplementederror

dinhlongviolin1 (approved):
lgtm

Issue #375, opened 2022-08-25
BUG-Date Selector / Localization Provider


FredLL-Avaiga:
**Description**
MUI: Can not find utils in context. It looks like you forgot to wrap your component in LocalizationProvider, or pass dateAdapter prop directly.

**How to reproduce**
Open a page with DateSelector

**Expected behavior**
Shouldn't error
An Error showing on the page at least

FredLL-Avaiga:
depends on the way we import it:
`import { LocalizationProvider } from "@mui/x-date-pickers/LocalizationProvider";` is valid
while `import { LocalizationProvider } from "@mui/x-date-pickers";` does not provoke any 'compilation' error but generates a runtime error :-(

FredLL-Avaiga:
A good occasion to add a ErrorBoundary component in the render tree so that in case of error during the react render the user gets some information in the rendered page (instead of a completely white page)

Issue #265, opened 2022-05-17
Add /status page


florian-vuillemot:
**What would that feature address**
Add monitoring facility for external tools as sonde. Calling "/" of an application can have side effects (computation, db connection, etc.).

***Description of the ideal solution***
Application monitoring uses to call a "/static" page on an application to know if the application is still running. Information like "websocket are up" is a plus.

***Caveats***
Users can want to provide other information on this page for internal needs. Taipy can not handle this custom behaviour but should allow users to do it.
Ex: Ensure that my application can still communicate with an external tool. 

**Acceptance Criteria**
- [ ] This page should be resource-efficient. 
- [ ] Doc on the page for dev and operator.
- [ ] Doc for customization of this page.

FredLL-Avaiga:
![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/1344fc0f-15a2-4d31-b7b7-7e6a77ddf561)

What do you think @florian-vuillemot ?

FredLL-Avaiga:
After a enlightened comment from @FabienLelaquais 
the status.html page is now a status.json 
```
{
  "backend_version": "1.2.0", 
  "flask_version": "2.1.0", 
  "frontend_version": "1.2.0", 
  "host": "127.0.0.1:5000", 
  "python_version": "3.10.1 (tags/v3.10.1:2cd268a, Dec  6 2021, 19:10:37) [MSC v.1929 64 bit (AMD64)]", 
  "user_status": "on_status(state: State)"
}
```
Which is way better !

FredLL-Avaiga:
URL is /taipy.status.json

FredLL-Avaiga:
CVE informations are not shown by default @florian-vuillemot

FredLL-Avaiga:
commanded by config extended_status: bool
Default is;
```
{
  "gui": {
    "user_status": ""
  }
}
```
extended is:
```
{
  "gui": {
    "backend_version": "1.2.0", 
    "flask_version": "2.1.0", 
    "frontend_build_date": "2022-08-26T09:03:10.419Z", 
    "frontend_build_mode": "development", 
    "frontend_version": "1.2.0", 
    "host": "127.0.0.1:5000", 
    "python_version": "3.10.1 (tags/v3.10.1:2cd268a, Dec  6 2021, 19:10:37) [MSC v.1929 64 bit (AMD64)]",
    "user_status": "on_status(state: State)"
  }
}
```

PR #376, opened 2022-08-25
Status page and LocatizationProvider fix (#265 #375)

FredLL-Avaiga:
- #265 and a status.html page
- #375 fix LocalizationProvider issue
- add an ErrorBoundary component to catch error during react tree render


src/taipy/gui/server.py
```diff
@@ -110,6 +111,8 @@ def my_index(path):
                     styles=styles,
                     version=version,
                 )
+            if path == "status.json":
```
FabienLelaquais: Do you think it would be safer to have a 'taipy.status' path instead?
That both better protects user-defined page names and allows for some extensions in the future...

FabienLelaquais (approved):
I like that a lot!

src/taipy/gui/server.py
FredLL-Avaiga: ok for taipy.status.json :-)

dinhlongviolin1 (approved):
As usual, well done!

Issue #362, opened 2022-08-16
BUG-Notebook on VS Code


FredLL-Avaiga:
**Description**
From @AlbVu :

There is a new post on Taipy Forum that may affect people who try to use taipy on VSCode notebook: 
"[taipy’s output collapse in visual studio code notebook](https://www.taipy.io/forums/topic/taipys-out-put-colapsed-in-visual-studio-code-notebook/)", August 15th, 2022

I was able to replicate the errors directly but couldn't find the root cause. I also found a few posts that this maybe a Conda problem, so I also conducted some tests: 

1. The problem happens on both Mac and Windows, but only when using Notebooks on Visual Studio Code. It runs perfectly fine on JupyterLab (JL).

2. I also created a new non-Conda environment and once again, the same result happens (errors when running on VSC, but not JL). This may suggest that the problem is related to VSC and not Conda-related.

3. Running the script on a notebook on pycharm is also fine (no errors).

**Runtime environment**
Notebook in VSCode

PR #377, opened 2022-08-26
fix notebook bug (#362)

dinhlongviolin1:
Fix Bug #362 
The problem is much simpler than I thought, the ipykernel used by vscode have the same --force-https argument that we have


src/taipy/gui/config.py
```diff
@@ -170,7 +172,7 @@ def _handle_argparse(self, args):  # pragma: no cover
             config["use_reloader"] = False
         if args.ngrok_token:
             config["ngrok_token"] = args.ngrok_token
-        if args.force_https:
+        if args.use_https:
             config["force_https"] = True
```
FredLL-Avaiga: Should we change the config param from force_https to use_https everywhere for the sake of uniformity and doc ?

FredLL-Avaiga (changes requested):

FredLL-Avaiga (approved):
I guess that https will be provided via a proxy ?

dinhlongviolin1:
> I guess that https will be provided via a proxy ?

Definitely

FabienLelaquais (approved):
Less is more.

PR #378, opened 2022-08-26
Cast value on table edit (#365)

FredLL-Avaiga:
on_edit payload receive now a user_value which is the string entered by the user
the value is casted to the type of the column


gui/doc/table.csv
```diff
@@ -47,7 +47,8 @@ This dictionary has the following keys:
 <ul>
 <li>index (int): the row index.</li>
 <li>col (str): the column name.</li>
-<li>value (any): the new cell value.</li>
+<li>value (any): the new cell value casted to the type of the column.</li>
```
FabienLelaquais: That would be 'cast', not 'casted'.

gui/src/components/Taipy/tableUtils.tsx
```diff
@@ -71,6 +71,9 @@ const formatValue = (val: RowValue, col: ColumnDesc, formatConf: FormatConfig, n
             return val as string;
     }
 };
+const VALID_BOOLEAN_STRINGS = ["true", "1", "t", "y", "yes", "yeah", "sure"];
```
FabienLelaquais, line 74: This is kind of similar to what is done in utils/boolean.py (where by the way, choices are duplicated, which is bad).
I don't see a way to share that choices array... if you do, please do.

FabienLelaquais (approved):
👌

gui/src/components/Taipy/tableUtils.tsx
FredLL-Avaiga: Copy/paste is the best I could achieve in this case ...
I think it would be way overkill to pass the array as a window variable when constructing the index.html from template !

dinhlongviolin1 (approved):
Simple one!

Issue #384, opened 2022-08-29
Use Decimator on a sub-area of a whole chart.


FabienLelaquais:
**What would that feature address**

Bla

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #390, opened 2022-08-31
chart zooming (#384)

dinhlongviolin1:
```python
import pandas as pd
import numpy as np
from src.taipy.gui import Gui
from src.taipy.gui.data.decimator import RDP

df = pd.DataFrame(np.random.randint(0, 100000, size=(30000, 1)), columns=["A"])
df["index1"] = df.index

decimator_instance = RDP(n_out=300)
# decimator_instance = RDP(n_out=300, chart_zooming=False)

gui = Gui(
    page=""" 
<|{df}|chart|x=index1|y=A|decimator=decimator_instance|>
"""
)
gui.run(run_browser=False)

```


gui/doc/chart.csv
```diff
@@ -13,14 +13,6 @@ text,indexed(str),,"Column name for the text associated to the point for the ind
 This is meaningful only when _mode_ has the <i>text</i> option."
 title,str,,The title of this chart control.
 render,bool,TRUE,"If True, this chart is visible on the page."
-on_range_change,Callback,,"Callback function called when the visible part of the x axis changes.
```
FabienLelaquais: Whao.
How come this doc is gone?

src/taipy/gui/data/pandas_data_accessor.py
```diff
@@ -270,17 +270,24 @@ def __get_data(  # noqa: C901
...
-                    warnings.warn(f"Limit rows error for dataframe: {e}")
+            if isinstance(decimator_instance, PropertyType.decimator.value):
+                x_column, y_column = payload.get("xAxis", ""), payload.get("yAxis", "")
+                if decimator_instance._chart_zooming and "relayoutData" in payload:
+                    chart_modes = payload.get("chartModes", [])
+                    relayoutData = payload.get("relayoutData", {})
+                    x0 = relayoutData.get("xaxis.range[0]")
+                    x1 = relayoutData.get("xaxis.range[1]")
+                    y0 = relayoutData.get("yaxis.range[0]")
+                    y1 = relayoutData.get("yaxis.range[0]")
```
FabienLelaquais: Are you sure that is [0]?

FabienLelaquais (changes requested):
All this looks excellent!
I'm quite worried about the range_change drop though. This is not only very useful, but genuinely part of the product...

gui/src/components/Taipy/Chart.tsx
```diff
@@ -212,12 +209,16 @@ const Chart = (props: ChartProp) => {
     useEffect(() => {
         if (refresh || !data[dataKey.current]) {
             const backCols = Object.keys(config.columns).map((col) => config.columns[col].dfid);
+            const xAxis = layout.xaxis.title instanceof Object && "text" in layout.xaxis.title ? layout.xaxis.title.text : layout.xaxis.title;
```
FredLL-Avaiga: Why not use 
```
const xAxis = config.traces.length && config.traces[0].length && config.traces[0][0] && config.columns[config.traces[0][0]].dfid;
const yAxis = config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] &&  config.columns[config.traces[0][1]].dfid;
```
That would give you the real column names (and not the titles associated to the axis)

gui/src/components/Taipy/Chart.tsx
```diff
@@ -212,12 +209,16 @@ const Chart = (props: ChartProp) => {
...
+            const yAxis = layout.yaxis.title instanceof Object && "text" in layout.yaxis.title ? layout.yaxis.title.text : layout.yaxis.title;
```
FredLL-Avaiga: I guess we could also define a "sub-payload" for decimator so that we don't calculate things if there's no decimator
ie 
```
const decimatorPayload = decimator ? {
  width: plotRef.current?.clientWidth,
   yAxis: config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
   decimator: decimator
} : undefined;
```
Then pass this as a decimator parameter to createRequestChartUpdateAction ?

gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: And we need to update the dependency list with config.columns & config.traces

gui/src/components/Taipy/Chart.tsx
```diff
@@ -355,9 +356,30 @@ const Chart = (props: ChartProp) => {
     }, [active, props.plotConfig]);
 
     const onRelayout = useCallback(
-        (eventData: PlotRelayoutEvent) =>
-            tp_onRangeChange && dispatch(createSendActionNameAction(id, { action: tp_onRangeChange, ...eventData })),
```
FredLL-Avaiga, line 359: Should we not keep the on_range_change for other uses (when there's no decimator) ?

gui/src/context/taipyReducers.ts
```diff
@@ -21,6 +21,7 @@ import { TIMEZONE_CLIENT } from "../utils";
 import { parseData } from "../utils/dataFormat";
 import { MenuProps } from "../utils/lov";
 import { FilterDesc } from "../components/Taipy/TableFilter";
+import { PlotRelayoutEvent } from "plotly.js";
```
FredLL-Avaiga: I don't think we want/need this dependency here

gui/src/context/taipyReducers.ts
```diff
@@ -476,18 +477,26 @@ export const createRequestChartUpdateAction = (
     name: string | undefined,
     id: string | undefined,
     columns: string[],
+    xAxis: unknown,
```
FredLL-Avaiga: use a decimator object ?

src/taipy/gui/renderers/factory.py
```diff
@@ -94,7 +94,6 @@ class _Factory:
                 ("height", PropertyType.string_or_number),
                 ("layout", PropertyType.dict),
                 ("plot_config", PropertyType.dict),
-                ("on_range_change", PropertyType.function),
```
FredLL-Avaiga: discuss ?

FredLL-Avaiga (changes requested):

gui/src/components/Taipy/Chart.tsx
```diff
@@ -355,9 +360,32 @@ const Chart = (props: ChartProp) => {
...
+                }
+                dispatch(
+                    createRequestChartUpdateAction(
+                        updateVarName,
+                        id,
+                        backCols,
+                        dataKey.current,
+                        decimatorPayload,
+                        config.modes,
+                        eventData,
```
FredLL-Avaiga: I thought that eventData and config.modes might be part of decimatorPayload ?

FredLL-Avaiga (changes requested):

gui/src/context/taipyReducers.ts
```diff
@@ -477,17 +477,15 @@ export const createRequestChartUpdateAction = (
     id: string | undefined,
     columns: string[],
     pageKey: string,
-    width?: number,
-    decimator?: string
+    decimatorPayload?: unknown | undefined,
```
FredLL-Avaiga: I thought that `decimatorPayload?: unknown` was equal to `decimator: unknown | undefined` ?

FredLL-Avaiga (approved):
NIce work

gui/src/context/taipyReducers.ts
dinhlongviolin1: You are right!

FabienLelaquais (approved):
Looking really good.
Can't wait to see that in action in our user's apps!

Issue #363, opened 2022-08-17
Update Flask to 2.2


dinhlongviolin1:
**Description**
Right now there is some incompatibility issues with Flask 2.2. 
One side thing, Flask dev server can be removed and use gevent exclusively in notebook environment

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #395, opened 2022-09-05
upgrade flask to 2.2 (#363)

dinhlongviolin1:
- Upgrade Flask to 2.2
- Add flask new Json Provider class
- Make gevent default for all use case including notebook
- make simple-websocket optional


Pipfile
```diff
@@ -22,7 +22,6 @@ gevent = "==21.12.0"
 gevent-websocket = "==0.10.1"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
 kthread = "==0.2.3"
-werkzeug = "==2.0.3"
```
FredLL-Avaiga, line 25: I suppose the right version of werzeug is pulled by flask >= 2.2 ?

setup.py
```diff
@@ -55,7 +53,8 @@
         "python-magic>=0.4.24,<0.5;platform_system!='Windows'",
         "python-magic-bin>=0.4.14,<0.5;platform_system=='Windows'",
     ],
-    "arrow": ["pyarrow>=7.0,<9.0"],
+    "arrow": ["pyarrow>=9.0,<10.0"],
```
FredLL-Avaiga, line 56: quietly updating py-arrow ?

FabienLelaquais (approved):
All this looks excellent to me. Well done!

tox.ini
```diff
@@ -28,8 +28,6 @@ commands =
 platform = linux
 commands =
     pipenv install --dev --skip-lock
-    pipenv run pip install "flask==2.1"
```
FredLL-Avaiga, line 31: less is more 👍

FredLL-Avaiga (approved):

Pipfile
dinhlongviolin1: definitely. the only reason why we need that previously was to run notebook on flask built in server but since we are using gevent for all use case now, this is not needed anymore

setup.py
dinhlongviolin1: You did update that in the frontend so I am going to update it in the backend :D

Issue #405, opened 2022-09-10
BUG-Extension Wrong variable name


FredLL-Avaiga:
**Description**
The variable name passed to the extension library method get_data is the user variable name but the taipy calculated variable name should be returned in the payload

**How to reproduce**
Try to use a data attribute in an extension libray element

**Expected behavior**
Pass the user variable name to the library but keep the Taipy calculated variable name internally

PR #406, opened 2022-09-10
#405 user variable name vs var_name

FredLL-Avaiga:
Visible only with data using element from extension library and no other impact


FabienLelaquais (approved):
🤔

dinhlongviolin1 (approved):

Issue #410, opened 2022-09-13
BUG-Cant start gui while in debug mode with taipy-enterprise


dinhlongviolin1:
**Description**
Issue with running taipy-enterprise in vscode debug mode

**How to reproduce**
A code fragment
```
from taipy.gui import Gui
gui = Gui("test")
print(gui)
gui.run(run_browser=False, async_mode="threading")
...
```
...

**Expected behavior**
Description of what would be the expected outcome.

**Screenshots**
When available and relevant, screenshots better help show the problem.

**Runtime environment**
Please specify relevant indications.
 - OS: [e.g. Linux, Windows] and version
 - Browser: [e.g. Chrome, Edge, Safari] and version
and any other relevant information.

PR #411, opened 2022-09-13
access locals safely (#410)

dinhlongviolin1:
Fix Bug #410 per FJ


FredLL-Avaiga (approved):
Good catch.
I gues we should always use dict.get() instead of dict[]

FabienLelaquais (approved):
🤨well done

Issue #412, opened 2022-09-13
Extension API: Hide Context and Store as long as possible.


FabienLelaquais:
**Description**

The communication between the front-end side of the components to the back-end elements is performed using the `useContext()` React hook, that uses the Taipy Context to handle the Taipy Store object.

This is pretty much mandatory for each and every extended element, forcing programmers to call:<br/>
    `const { dispatch } = useContext<Store>(Context);`
then use `dispatch` to feed the backend.

A more simple way to do that would be to provide a predefined hook (`useDispatch()`?) that would hide both `Context` and `Store`, so long as users are not exposed to these classes:<br/>
    `const dispatch = useDispatch();`

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #413, opened 2022-09-13
#412 useDispatch

FredLL-Avaiga:
publish useDispatch which is a shortcut for const {dispatch} = useContext<TaipyStore>(TaipyContext)

### DIFF
--- a/gui/packaging/taipy-gui.d.ts
+++ b/gui/packaging/taipy-gui.d.ts
@@ -11,6 +11,8 @@
  * specific language governing permissions and limitations under the License.
  */
 
+import * as React from "react";
+
 /**
  * This extracts the backend name of a given property identified by `name`.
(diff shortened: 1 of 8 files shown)


florian-vuillemot (approved):
LGTM

gui/packaging/taipy-gui.d.ts
```diff
@@ -11,6 +11,8 @@
  * specific language governing permissions and limitations under the License.
  */
 
+import * as React from "react";
```
FabienLelaquais, line 14: I don't believe you wrote that

FabienLelaquais (approved):
Good move

dinhlongviolin1 (approved):

gui/packaging/taipy-gui.d.ts
FredLL-Avaiga: I did

PR #414, opened 2022-09-14
decimator improvement

dinhlongviolin1:
- decimate by chart types
- get chart decimate nb_rows

### DIFF
--- a/gui/src/components/Router.tsx
+++ b/gui/src/components/Router.tsx
@@ -111,7 +111,7 @@ const Router = () => {
                                     </Box>
                                     {state.ackList.length ? (
                                         <Box sx={progressSx}>
-                                            <CircularProgress size="1em" />
+                                            <CircularProgress size="1em" disableShrink/>
                                         </Box>
                                     ) : null}
                                 </Box>
--- a/gui/src/components/Taipy/Chart.tsx
+++ b/gui/src/components/Taipy/Chart.tsx
@@ -220,7 +220,7 @@ const Chart = (props: ChartProp) => {
                 xAxis: config.traces.length && config.traces[0].length && config.traces[0][0] && config.columns[config.traces[0][0]].dfid,
                 yAxis: config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
                 decimator: decimator,
-                chartModes: config.modes,
+                chartMode: config.modes[0],
               } : undefined;
             dataKey.current = backCols.join("-") + (decimator ? `--${decimator}` : "");
             dispatch(
@@ -377,7 +377,7 @@ const Chart = (props: ChartProp) => {
                     yAxis: config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
                     decimator: decimator,
                     relayoutData: eventData,
-                    chartModes: config.modes,
+                    chartMode: config.modes[0],
                 }
                 dispatch(
                     createRequestChartUpdateAction(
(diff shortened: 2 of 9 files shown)


gui/src/components/Taipy/Chart.tsx
```diff
@@ -219,7 +219,8 @@ const Chart = (props: ChartProp) => {
                 height: plotRef.current?.clientHeight,
                 xAxis: config.traces.length && config.traces[0].length && config.traces[0][0] && config.columns[config.traces[0][0]].dfid,
                 yAxis: config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
-                decimator: decimator
+                decimator: decimator,
+                chartModes: config.modes,
```
FredLL-Avaiga: I suppose you only need config.modes[0] to be consistent with config.traces[0] used for xAxix & yAxis ?

src/taipy/gui/data/utils.py
```diff
@@ -21,13 +22,19 @@
...
+    _CHART_MODES: t.List[str] = []
+
     def __init__(self, threshold: t.Optional[int], zoom: t.Optional[bool]) -> None:
         """TODO: Decimator class description"""
         super().__init__()
         self.threshold = threshold
         self._zoom = zoom if zoom is not None else True
 
-    def _is_applicable(self, data: t.Any, nb_rows_max: int):
+    def _is_applicable(self, data: t.Any, nb_rows_max: int, chart_modes: t.List[str]):
```
FredLL-Avaiga: then chart_modes => chart_mode: str

src/taipy/gui/gui.py
```diff
@@ -472,7 +472,7 @@ def __get_real_var_name(self, var_name: str) -> t.Tuple[str, str]:
                 raise NameError(f"Can't find matching variable for {var_name} on context: {current_context}")
         return f"{var_name}.{suffix_var_name}" if suffix_var_name else var_name, current_context
 
-    def __call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[str] = None):
+    def _call_on_change(self, var_name: str, value: t.Any, on_change: t.Optional[str] = None):
         try:
```
FredLL-Avaiga, line 476: Will we get an exception in __get_real_var_name with <chart var name>.decimator.... ?

gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: And I suppose we need to write somewhere in the doc that the decimator only applies to the first trace of the chart (at least for now) ?
Or handle the case to use decimator in other traces ?

FredLL-Avaiga (changes requested):
I suppose we'll force the decimator to be used only on the 1st trace for now ?

src/taipy/gui/gui.py
dinhlongviolin1: I believe we won't get an exception since the table_var_name is valid and will get discovered by the get_real_var_name method

gui/src/components/Taipy/Chart.tsx
dinhlongviolin1: I agree, we will only get the first traces (main data) for now

gui/src/components/Router.tsx
```diff
@@ -111,7 +111,7 @@ const Router = () => {
                                     </Box>
                                     {state.ackList.length ? (
                                         <Box sx={progressSx}>
-                                            <CircularProgress size="1em" />
+                                            <CircularProgress size="1em" disableShrink/>
```
FredLL-Avaiga, line 114: Why ?

FredLL-Avaiga (approved):

FabienLelaquais (approved):
Excellent idea!

Issue #416, opened 2022-09-19
BUG-No reloading after the user changes their files 


FlorianJacta:
**Description**
Taipy GUI (1.1.2 and 1.1.3) doesn't reload the entire script when a file is changed in the project. The execution has to be killed and re-run. Taipy GUI 1.1.1 works fine.

**How to reproduce**
- Execute this code
```
from taipy.gui import Gui

md = """
test
"""

Gui(md).run(use_reloader=True)
```
- Change the file. For example, change _md_. The script is not re-run and no changes can be seen on the client.

**Expected behavior**
Taipy should reload the entire script when a file is modified.

PR #417, opened 2022-09-19
Async mode set to threading in debug mode (#416)

dinhlongviolin1:
Undirectly solve issue #416. Userman will need to be updated to document the change.


src/taipy/gui/gui.py
```diff
@@ -1493,6 +1493,9 @@ def run(
         # Load application config from multiple sources (env files, kwargs, command line)
         self._config._build_config(run_root_dir, self.__env_filename, kwargs)
 
+        if app_config["debug"]:
+            async_mode = "threading"
```
FredLL-Avaiga, line 1497: Add a warning if async_mode is not threading?

FredLL-Avaiga (approved):
Simple enough

src/taipy/gui/gui.py
FabienLelaquais: You mean if it is forced to something else?
That would make sense...

FabienLelaquais (approved):
Good to go!

PR #427, opened 2022-09-29
add is debugging method (#428)

dinhlongviolin1:
Solve issue #428


FabienLelaquais (approved):
Good!
Now one day maybe we'll expose this at the 'taipy' level...

FredLL-Avaiga (approved):
I thought we'd set the Debug config flag if it is not explicitly set ?

Issue #430, opened 2022-09-29
Add a template directory structure for dynamic custom visual elements


FabienLelaquais:
**What would that feature address**

Getting started on the extension API, and specifically with dynamic custom elements (that need JavaScript code) is
not that obvious.

We want to add to the distribution an entire directory structure that demonstrates a very basic yet dynamic control,
and explain all it takes to build and package this custom library.

***Description of the ideal solution***

A directory that show how a custom library containing a dynamic element is available for the developer, as well as it can be referenced from the User Manual.

**Acceptance Criteria**
- [ ] Ensure code is tested.
- [ ] Use that directory from the User Manual

PR #434, opened 2022-09-30
First version of element library  template (#430)

FabienLelaquais:
Addresses #430.


doc/extension/Pipfile
```diff
@@ -0,0 +1,12 @@
...
+verify_ssl = true
+name = "pypi"
+
+[packages]
+taipy-gui = ">=2.0"
+
+[dev-packages]
+
+[requires]
+python_version = "3"
```
FredLL-Avaiga, line 12: `>= 3.8` ?

doc/extension/README.md
```diff
@@ -0,0 +1,171 @@
...
+- If you are using `pipenv`:
+   ```sh
+   $ pipenv --python $PYTHON_VERSION
+   $ pipenv shell
+   $ pipenv install
+   ```
+
+- If you are not using `pipenv`:
+   ```sh
+   $ python -m venv ./venv
```
FredLL-Avaiga, line 76: Where does venv come from?

FredLL-Avaiga (approved):
Don't forget to rename js to webui

doc/extension/README.md
FabienLelaquais: Looks like it comes out of the box with Python... who would have guessed?

doc/extension/Pipfile
FabienLelaquais: I guess that makes a lot of sense. Hopefully this syntax is allowed...

FabienLelaquais: It is not. Only a specific version (3.8, 3.9...) is.
Therefore... 3 works here

dinhlongviolin1 (approved):
Nicely Done!

FredLL-Avaiga (approved):
go

PR #439, opened 2022-10-03
Support decimator for 3d chart (#440)

dinhlongviolin1:
Solve Issue #440

```python
import pandas as pd
import numpy as np
from src.taipy.gui import Gui
from src.taipy.gui.data.decimator import ScatterDecimator

from sklearn.datasets import make_classification

n_clusters = 3
n_features = 5

X, y = make_classification(
    n_samples=10000,
    n_features=n_features,
    n_informative=n_features,
    n_repeated=0,
    n_redundant=0,
    n_classes=n_clusters,
    class_sep=n_clusters,
)
df = pd.DataFrame(X[:, :3], columns=["x", "y", "z"])
df["class"] = y

decimator = ScatterDecimator(binning_ratio=50, max_overlap_points=1)

def create_colors(y):
    colors = []
    for i in range(len(y)):
        if y[i] == 0:
            colors += ["red"]
        if y[i] == 1:
            colors += ["blue"]
        if y[i] == 2:
            colors += ["green"]
    return colors

def on_change(state, var, val):
    print(var, val)

df["Colors"] = create_colors(y)

marker = {"color": "Colors"}

chart = """
<|layout|

<|{df}|chart|type=Scatter3D|x=x|y=y|z=z|marker={marker}|mode=markers|height=800px|decimator=decimator|>

<|{df}|chart|type=Scatter3D|x=x|y=y|z=z|marker={marker}|mode=markers|height=800px|>

|>
"""


Gui(page=chart).run()

```


gui/src/components/Taipy/Chart.tsx
```diff
@@ -367,15 +368,16 @@ const Chart = (props: ChartProp) => {
...
                 const eventDataKey = Object.keys(eventData).map(v => v + "=" + eventData[v as keyof typeof eventData]).join("-");
                 dataKey.current = backCols.join("-") + (decimator ? `--${decimator}` : "") + "--" + eventDataKey;
                 const decimatorPayload = {
                     width: plotRef.current?.clientWidth,
                     height: plotRef.current?.clientHeight,
                     xAxis: config.traces.length && config.traces[0].length && config.traces[0][0] && config.columns[config.traces[0][0]].dfid,
-                    yAxis: config.traces.length == 1 && config.traces[0].length > 1 && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
+                    yAxis: config.traces.length > 0 && config.traces[0].length > 1 && config.traces[0][1] && config.columns[config.traces[0][1]] && config.columns[config.traces[0][1]].dfid,
+                    zAxis: config.traces.length > 0 && config.traces[0].length > 2 && config.traces[0][2] && config.columns[config.traces[0][2]] && config.columns[config.traces[0][2]].dfid,
```
FredLL-Avaiga: I guess we should extract this as a function 
`const getAxis = (traces, columns, index) => traces.length && traces[0].length > index && traces[0][index] && columns[traces[0][index]] && columns[traces[0][index]].dfid`

FredLL-Avaiga: and maybe even the complete decimator payload creation ...

FredLL-Avaiga (approved):

gui/src/components/Taipy/Chart.tsx
dinhlongviolin1: Thanks!

gui/src/components/Taipy/Chart.tsx
```diff
@@ -118,6 +118,38 @@ const getValueFromCol = (values: TraceValueType | undefined, col: string): (stri
     return [];
 };
 
+const getDecimatorPayload = (
+    decimator: string | undefined,
+    plotRef: React.RefObject<HTMLDivElement>,
+    modes: string[],
+    columns: Record<string, ColumnDesc>,
+    traces: string[][],
+    relayoutData: PlotRelayoutEvent | undefined = undefined
```
FredLL-Avaiga: could use the form relayoutData?: PlotRelayoutEvent :-)
if ommitted, the value is undefined

gui/src/components/Taipy/Chart.tsx
```diff
@@ -118,6 +118,38 @@ const getValueFromCol = (values: TraceValueType | undefined, col: string): (stri
     return [];
 };
 
+const getDecimatorPayload = (
+    decimator: string | undefined,
+    plotRef: React.RefObject<HTMLDivElement>,
```
FredLL-Avaiga: could have passed plotRef.current as plotDiv?: HTMLDivElement

gui/src/components/Taipy/Chart.tsx
```diff
@@ -118,6 +118,38 @@ const getValueFromCol = (values: TraceValueType | undefined, col: string): (stri
...
+    modes: string[],
+    columns: Record<string, ColumnDesc>,
+    traces: string[][],
+    relayoutData: PlotRelayoutEvent | undefined = undefined
+) => {
+    return decimator
+        ? {
+              width: plotRef.current?.clientWidth,
+              height: plotRef.current?.clientHeight,
+              xAxis: traces.length && traces[0].length && traces[0][0] && columns[traces[0][0]].dfid,
```
FredLL-Avaiga: I would still make a function of this (3 calls)
And this way the value can be set to `undefined` instead of `false` when there's no data for the axis

FredLL-Avaiga (approved):
a few more comments
Well done

gui/src/components/Taipy/Chart.tsx
```diff
@@ -127,6 +127,34 @@ const getValueFromCol = (values: TraceValueType | undefined, col: string): (stri
...
 
+const getAxis = (traces: string[][], columns: Record<string, ColumnDesc>, axis: number) => {
+    if (traces.length > 0 && traces[0].length > axis && traces[0][axis] && columns[traces[0][axis]])
+        return columns[traces[0][axis]].dfid
+    return undefined
+}
+
+const getDecimatorPayload = (
+    decimator: string | undefined,
+    plotDiv: HTMLDivElement | null,
```
FabienLelaquais, line 138: Is this equivalent to `plotDiv?: HTMLDivElement,`
that we use in most places?

FabienLelaquais (approved):
Impressive!

gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: decimator: string | undefined is equivalent to decimator?: string
not sure we need to manage null vs undefined in this case ?
I guess it's made necessary by the ref decl : const plotRef = useRef<HTMLDivElement>(null);
And I guess we can change this to const plotRef = useRef<HTMLDivElement>();

dinhlongviolin1: if strictNullChecks, then null option should be handled

dinhlongviolin1: but since we dont have strictNullChecks in tsconfig, i suppose handling null case is not necessary

PR #451, opened 2022-10-07
Build/#450 missing doc extension file

FabienLelaquais:
The fix is to manually force demo.py to be handled by git, in a local (doc/extension) .gitignore file.


FabienLelaquais:
Problem with last commit... creating another PR.

Issue #461, opened 2022-10-20
Extension: library should provide a way to specify an external resource (Absolute URL)


FredLL-Avaiga:
**Description**
Say you want to use a js library available via a CDN (for example `https://cdnjs.cloudflare.com/ajax/libs/frappe-gantt/0.6.1/frappe-gantt.min.js`)

We should provide a way to set in the html:
`<script src="https://cdnjs.cloudflare.com/ajax/libs/frappe-gantt/0.6.1/frappe-gantt.min.js" integrity="sha512-HyGTvFEibBWxuZkDsE2wmy0VQ0JRirYgGieHp0pUmmwyrcFkAbn55kZrSXzCgKga04SIti5jZQVjbTSzFpzMlg==" crossorigin="anonymous" referrerpolicy="no-referrer"></script>`

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
checking  hostname via urlparse is good enough I guess

PR #462, opened 2022-10-20
#461 support absolute URLS in library.get_scripts and get_styles

FredLL-Avaiga:
Will need a bit of doc I guess :
get_scripts
"""
Returns the list of resources names for the scripts.
if a resource name is an absolute URL it will be used as is, if it's not it will be passed to get_resource to retrieve a local Path to the resource
...


FabienLelaquais (approved):
But why didn't you integrate thedoc then?

dinhlongviolin1 (approved):
Well done as always!

FredLL-Avaiga:
> But why didn't you integrate thedoc then?

true ...

Issue #460, opened 2022-10-20
Migrate the Visual Elements documentation to JSON


FabienLelaquais:
**Description**

We will need, for Taipy Studio, to extract the elements' properties to provide autocompletion.
The current CSV format really is not handy.,

**Acceptance Criteria**
- [ ] Propagate the change to taipy-doc for the User Manual generation.

PR #468, opened 2022-10-24
Convert CSV file for elements doc to JSON (#460)

FabienLelaquais:
Addresses #460.

The Python code to do the job is the following:
```
import json
import numpy
import pandas as pd
import sys

for in_name in sys.argv[1:]:
  if in_name.endswith(".csv"):
    out_name = in_name[:len(in_name)-4]
  else:
    out_name = in_name
    in_name += ".csv"
  out_name += ".json"
  df = pd.read_csv(in_name, encoding="utf-8")
  inherits = []
  default_prop = None
  properties = []
  for row in list(df.to_records(index=False)):
    prop_name = row[0]
    if prop_name[0] == "*":
      if default_prop:
        raise ValueError(f"Multiple default property in {in_name}")
      prop_name = prop_name[1:]
      default_prop = prop_name
    elif prop_name[0] == ">":
      inherits.append(prop_name[1:])
      prop_name = None
    if prop_name:
      # This weird == test is to detect NaNs
      default_value = row[2] if row[2] == row[2] else None
      if type(default_value) is numpy.bool_:
        default_value = bool(default_value)
      properties.append({
        "name": prop_name,
        "type": row[1],
        "default": default_value,
        "doc": row[3]
      })
  element = {
    "default": default_prop
  }
  if inherits:
    element["inherits"] = inherits
  if properties:
    element["properties"] = properties
  with open(out_name, "w") as output_file:
    output_file.write(json.dumps(element, indent=2))
    print(f"{out_name} written.")
```


FredLL-Avaiga (approved):
I thought we'd have a unique json fille

FabienLelaquais:
> 

That would make inheritance tricky to code, wouldn't it?
If you think that Taipy Studio would benefit from a unique file, that would be pretty easy (and simplify the doc generation process...).
Please let me know. I can certainly change that.

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
+{
+  "controls": [
+    [
+      "text",
```
FredLL-Avaiga, line 4: the construct [[name, description]] does not garantee the unicity. What's the rational behind that choice ?

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+      {
+        "default": "value",
```
FredLL-Avaiga: Why not mark the property as default: true ?

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+            "doc": "Column name for the <i>base</i> value. Used in bar charts only.\nSee the Plotly <a href=\"https://plotly.com/javascript/reference/bar/#bar-base\">bar chart base</a> documentation for details.\""
+          },
+          {
+            "name": "title",
+            "type": "str",
+            "doc": "The title of this chart control."
+          },
+          {
+            "name": "render",
+            "type": "bool",
```
FredLL-Avaiga: should be dynamic(bool)

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+      {
+        "default": "render",
+        "inherits": [
+          "partial",
+          "shared"
+        ],
+        "properties": [
+          {
+            "name": "render",
+            "type": "bool",
```
FredLL-Avaiga: dynamic(bool)

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+      {
+        "default": "value",
+        "inherits": [
+          "partial",
+          "shared"
+        ],
+        "properties": [
+          {
+            "name": "value",
+            "type": "str",
```
FredLL-Avaiga: dynamic

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+            "doc": "Title of this block element."
+          },
+          {
+            "name": "expanded",
+            "type": "bool",
```
FredLL-Avaiga: dynamic

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+        "inherits": [
+          "partial",
+          "on_change",
+          "active",
+          "shared"
+        ],
+        "properties": [
+          {
+            "name": "open",
+            "type": "bool",
```
FredLL-Avaiga: dynamic

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+            "name": "persistent",
+            "type": "bool",
+            "default": false,
+            "doc": "If True, the pane appears next to the page.\nIf False, the pane covers the page where it appeared."
+          }
+        ]
+      }
+    ]
+  ],
+  "others": [
```
FredLL-Avaiga: naming ?

gui/doc/viselements.json
```diff
@@ -0,0 +1,1252 @@
...
+  "controls": [
+    [
+      "text",
+      {
+        "default": "value",
+        "inherits": [
+          "on_change",
+          "shared"
+        ],
+        "properties": [
```
FredLL-Avaiga, line 10: properties: { name: {... details}} would insure unicity

FredLL-Avaiga (approved):
Is the structure fixed by the doc requirements ?

gui/doc/viselements.json
FabienLelaquais: But I need to keep the order...

FredLL-Avaiga: I see : order or unicity, you chose
Another choice would be to add an index property but it's a pain ...

FabienLelaquais:
> 

No but the information has to be somewhere.
With this new structure, the order of elements controls the order they appear in the doc.

gui/doc/viselements.json
FabienLelaquais: I like that!

gui/doc/viselements.json
FabienLelaquais: 👍

gui/doc/viselements.json
FabienLelaquais: 👍

gui/doc/viselements.json
FabienLelaquais: 👍

gui/doc/viselements.json
FabienLelaquais: 👍

gui/doc/viselements.json
FabienLelaquais: Correct. Another requirement is that it should be quite easy to update...

gui/doc/viselements.json
FabienLelaquais: -> "undocumented"

gui/doc/viselements.json
FabienLelaquais: 👍

FredLL-Avaiga (approved):
No order needed for properties then ?

FabienLelaquais:
> No order needed for properties then ?

Yes yes! Order is required in the viselements.json file: this is used when generating the properties section for each element.

PR #476, opened 2022-10-28
'data' property is mandatory in the 'chart' control

FabienLelaquais:
No associated issue.


FredLL-Avaiga (approved):
Is it the only mandatory property ? ;-)
do you prefer that over required ?

FabienLelaquais:
> Is it the only mandatory property ? ;-)
Yes

> do you prefer that over required ?
No

Issue #474, opened 2022-10-27
Add polar types to charts


FlorianJacta:
**Description**
The goal is to open the API to include [polar charts](https://plotly.com/javascript/polar-chart/) in Taipy. _r_ and _theta_ are the parameters that need to be provided.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
Trying to be a bit more generic (ie support lat/lon as well)

FredLL-Avaiga:
r => x => r
theta => y => theta
for type == scatterpolar

FredLL-Avaiga:
scatterpolar & scatterpolargl are now supported by adding the alias r for x & theta for y.

Ex: 
```
from taipy.gui import Gui
import pandas as pd

data = [
  pd.DataFrame({
    "r": [0, 1.5, 1.5, 0, 2.5, 2.5, 0],
    "theta": [0, 10, 25, 0, 205, 215, 0]}), 
  pd.DataFrame({
    "r": [0, 3.5, 3.5, 0],
    "theta": [0, 55, 75, 0]}), 
  pd.DataFrame({
    "r": [0, 4.5, 4.5, 0, 4.5, 4.5, 0],
    "theta": [0, 100, 120, 0, 305, 320, 0]}), 
  pd.DataFrame({
    "r": [0, 4, 4, 0],
    "theta": [0, 165, 195, 0]}), 
  pd.DataFrame({
    "r": [0, 3, 3, 0],
    "theta": [0, 262.5, 277.5, 0]}),
]

options1 = {"fill": "toself", "fillcolor": "#709BFF"}
options2 = {"fill": "toself", "fillcolor": "#E4FF87"}
options3 = {"fill": "toself", "fillcolor": "#FFAA70"}
options4 = {"fill": "toself", "fillcolor": "#FFDF70"}
options5 = {"fill": "toself", "fillcolor": "#B6FFB4"}
line = {"color": "black"}

layout1 = {
    "polar": {
        "radialaxis": {
            "visible": True,
            "range": [0, 5]
        }
    },
    "showlegend": False
}

md = """
# Area Polar Chart

[...]

PR #479, opened 2022-10-31
polar charts (#474)

FredLL-Avaiga:
Added property r & theta which are aliases to x & y
Check issue #474 for examples

### DIFF
--- a/src/taipy/gui/renderers/builder.py
+++ b/src/taipy/gui/renderers/builder.py
@@ -60,6 +60,21 @@ class _Builder:
 
     __BLOCK_CONTROLS = ["dialog", "expandable", "pane", "part"]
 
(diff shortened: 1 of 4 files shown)


src/taipy/gui/renderers/builder.py
```diff
@@ -60,6 +60,21 @@ class _Builder:
 
     __BLOCK_CONTROLS = ["dialog", "expandable", "pane", "part"]
 
+    __CHART_AXYS_SUBSTITUTION = {
```
FabienLelaquais, line 63: So that IS a smart way to deal with these aliases. Well done!

FabienLelaquais (approved):
I love it!

dinhlongviolin1 (approved):
Simple enough for you 🥇

Issue #455, opened 2022-10-10
BUG- Wrong Axis Issue with histogram


vhfgosselin:
**Description**
The syntax to display a histogram chart is wrong.
We should set the x value currently we need to set the y value.

**How to reproduce**
A code fragment
```
import numpy as np
from taipy import Gui
import pandas as pd

list_r = np.random.randint(1,50,500)
df = pd.DataFrame(list_r,columns=['value'])
print(df)
page = '''
# first page
<|{df}|chart|type=histogram|y=value|>
'''

Gui(page=page).run()
```

**Expected behavior**
we just need to replace:
<|{df}|chart|type=histogram|y=value|>
with
<|{df}|chart|type=histogram|x=value|>

This is logical and consistent with plotly syntax: See https://plotly.com/python/histograms/

FredLL-Avaiga:
Did you try to not set y or x ?
<|{df}|chart|type=histogram|>

vhfgosselin:
Just did. Nothing appears, I get an empty chart.
Also, even if it were working, how would I plot 2 histograms on the same chart?

PR #481, opened 2022-11-02
#455 do not add an index axys on histogram

FredLL-Avaiga:


FabienLelaquais (approved):
Looks good to me... whatever that does!

dinhlongviolin1 (approved):

Issue #480, opened 2022-11-02
Changing class_name through callbacks not working


aazimh:
* Taipy version: 2.0.0
* Python version: 3.9.13
* Operating System: Windows

### Description

Trying to change the styling of a text block based on its value - the value is a percentage that is calculated in a callback, and based on how high/low it is I would like it to turn red/orange/green. 

### What I Did

This is the block for the actual text:

```MAPE is: <|{mape}|class_name={mape_class}|>```

This is the code that calculates the MAPE and updates its value within a callback:

```
mape_num = mean_absolute_percentage_error(temp_predicted['Actual'],temp_predicted['Predicted'])*100
if mape_num > 30:
    state.mape_class = 'mape_bad'
elif mape_num > 20:
    state.mape_class = 'mape_avg'
else:
    state.mape_class = 'mape_good'
state.mape = str(round(mape_num,1)) + "%" ### THIS SUCCESSFULLY UPDATES THE BLOCK ABOVE WITH THE RIGHT VALUE
```
These are the CSS classes:

```
<style>
.mape_good {
    color: green;
    font-weight: 1000;
    font-size: 1.5em;
    }
    
.mape_avg {
    color: orange;
    font-weight: 1000;
    font-size: 1.5em;
    }

.mape_bad {
    color: red;
    font-weight: 1000;
    font-size: 1.5em;
    }
</style>
```

Basically, the actual value updates successfully, but the class_name parameter does not and so it remains the default style. 

Curious to know if I'm doing something wrong or if there are other ways of applying different styles to elements based on conditions. Thank you!

FredLL-Avaiga:
Class_name is not a dynamic property so it won't be updated unless the page is refreshed.
There are a few workaround:
1/ Use a partial that would hold your text component and that you can refresh dynamically
2/ Use the new extension API to create your own text component with a dynamic style attribute

FredLL-Avaiga:
@aazimh This will be available in Taipy 2.1

aazimh:
@FredLL-Avaiga Wow, I didn't even get the time to explore partials and you already added it as a feature! Thank you so much - is there a place I can check expected version release dates? Want to know if I should wait for 2.1 or check partials out anyway.

FredLL-Avaiga:
Next version is due at the beginning of next year.

aazimh:
[...]

PR #484, opened 2022-11-07
Dynamic class name in visual elements (#480)

FredLL-Avaiga:
How to use ? the same as before
```

class_name = "red"
value = ""

def on_change(state: State, varname: str, value):
    if varname == "value":
        state.class_name = "green" if len(value) % 2 == 1 else "red"

page = """
# Dynamic class name

<|{value}|input|class_name={class_name}|>
"""


if __name__ == "__main__":
    Gui(page=page).run(async_mode="threading", use_relaoder=False)
```

### DIFF
--- a/gui/doc/viselements.json
+++ b/gui/doc/viselements.json
@@ -1254,7 +1254,7 @@
           },
           {
             "name": "class_name",
-            "type": "str",
+            "type": "dynamic(str)",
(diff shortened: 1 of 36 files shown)


FabienLelaquais (approved):
Whao nice job.
I'm sorry this impacted... well everything.

FabienLelaquais:
Did I miss the documentation update?

dinhlongviolin1 (approved):
Well done!

PR #489, opened 2022-11-15
inject config with documentation

jrobinAV:


src/taipy/gui/config.py
```diff
@@ -243,9 +244,20 @@ def _update(self, as_dict: t.Dict[str, t.Any]):
 
         @staticmethod
         def _configure(**properties):
+            """Configure GUI.
```
FabienLelaquais: -> "Configure the Graphial User Interface."

src/taipy/gui/config.py
```diff
@@ -243,9 +244,20 @@ def _update(self, as_dict: t.Dict[str, t.Any]):
...
+
+            Parameters:
+                **properties (Dict[str, Any]): Keyword arguments that configure `Gui` behavior.<br/>
```
FabienLelaquais: "... configure `Gui` behavior." -> "... configure the behavior of the \`Gui^\` instances."

src/taipy/gui/config.py
```diff
@@ -243,9 +244,20 @@ def _update(self, as_dict: t.Dict[str, t.Any]):
...
+                    Please refer to the
+                    [Configuration section](../gui/configuration.md#configuring-the-gui-instance)
+                    of the User Manual for more information on the accepted arguments.
+            Returns:
+                `_GuiSection^`: The gui configuration.
```
FabienLelaquais: Private class?

FabienLelaquais (approved):
Minor changes

src/taipy/gui/config.py
jrobinAV:
```suggestion
            """Configure the Graphical User Interface.
```

src/taipy/gui/config.py
jrobinAV:
```suggestion
                **properties (Dict[str, Any]): Keyword arguments that configure the behavior of the `Gui^` instances.<br/>
```

src/taipy/gui/config.py
jrobinAV: Yes, I am surprised as well. I did not want to change it since I guessed it was on purpose.

I would personally make it public and move it from a function definition so it is an outer class.

dinhlongviolin1 (approved):

src/taipy/gui/config.py
dinhlongviolin1: it can be easily moved to another file and conditionally imported in the __init__ file

dinhlongviolin1: I can do it within this PR if both of you agree

FabienLelaquais: Well why not...

jrobinAV: Great! 
Thx.

PR #495, opened 2022-11-18
Update yaml

FredLL-Avaiga:
- node version
- ubuntu LTS version
- actions version
- limit playwright version to not pull greenlet > 2


FabienLelaquais (approved):
Ubunto LTS is indeed a good idea.
Thx!

dinhlongviolin1 (approved):
Amazing work!

Issue #498, opened 2022-11-23
BUG-local ressources are accessible via the browser


FredLL-Avaiga:
**Description**
every file at the level of the main python file is accessible through de browser.

![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/b680dc63-c0d3-4e85-a174-e75f8f39549c)

**Expected behavior**
Provide a way to limit access
ie path_mapping ...
But default behavior remains the same

PR #499, opened 2022-11-23
Protect local files against unwanted access (#498)

FredLL-Avaiga:
by default, every resource is accessible through the browser.
If path_mapping is specified and not empty, access is restricted.

```
import numpy as np
from taipy.gui import Gui
import pandas as pd

list_r = np.random.randint(1,50,500)
df = pd.DataFrame(list_r,columns=['value'])
options={"nbinsx": 5}
page = '''
# first page
<|{df}|chart|type=histogram|options={options}|>
'''

Gui(page=page, path_mapping={"toto": "1"}).run(async_mode="threading")
```
image

While with 
`Gui(page=page).run(async_mode="threading")
`
or
`Gui(page=page, path_mapping={}).run(async_mode="threading")
`image

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -213,7 +213,9 @@ class Gui:
                 you want to access them using only '_assets_' in your application, you can
                 set _path_mapping={"assets": "/home/me/app_assets"}_. If your application
                 then requests the file _"/assets/images/logo.png"_, the server searches
-                for the file  _"/home/me/app_assets/images/logo.png"_.
+                for the file  _"/home/me/app_assets/images/logo.png"_.<br/>
+                If empty or not defined, access through the browser to all resources under the directory of the main python file
+                is allowed.
             env_filename (Optional[str]): An optional file from which to load application
                 configuration variables (see the
                 [Configuration](../gui/configuration.md#configuring-the-gui-instance) section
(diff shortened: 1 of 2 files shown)


FabienLelaquais:
I understand that it makes sense to re-use 'path_mapping' to do the job.
However, it's not a mapping anymore is it?

What's the meaning of the "1" value you have set in your example.
If this value is a string, the server will map the path to that value, right?

I guess maybe a more readable version of that would be a path, set to something clearly not a path. Like False.
And I would love to see the paths support patterns as well. Like "*.py". Does this make sense?

FredLL-Avaiga:
> I understand that it makes sense to re-use 'path_mapping' to do the job. However, it's not a mapping anymore is it?
> 
> What's the meaning of the "1" value you have set in your example. If this value is a string, the server will map the path to that value, right?
> 
> I guess maybe a more readable version of that would be a path, set to something clearly not a path. Like False. And I would love to see the paths support patterns as well. Like "*.py". Does this make sense?

I suppose this example is a bit too simple ?
path_mapping is still usable and I'm sure that user app will need it.
In this example it is only defined so that access is restricted 
but if a resource is requested with a starting path of "/toto/<rest of path to resource>", the resource "1/<rest of path to resource>" will be returned if the file exists.
I guess we could enhance the system to allow for glob filtering but that would be outside of path_mapping

dinhlongviolin1 (approved):
Nicely done!

FabienLelaquais (approved):
That will do just fine... for the time being.

Issue #458, opened 2022-10-12
BUG-notify fails to appear


mshell99:
**Description**
notification fails to appear in browser. notify is called twice in callback which otherwise is proeprely executed. No errors/warings in onsole.

**How to reproduce**
Full application code attached here.  Build app from requirements.txt.  Create a scenario (default settings OK). Go to "orders" page and click the "run scenarios" button at bottom of page.  If CPlex is not installed, comment out line 103 of main.py (call to solve) 

**Expected behavior**
2 notifications should appear, one at start of optimization one at end
**Screenshots**
[Shopping.zip](https://github.com/Avaiga/taipy-gui/files/9766259/Shopping.zip)

mshell99:
When I run this in Ubuntu VM,, second notification appears but not first

FredLL-Avaiga:
Can you share a piece of code ?
Are youusing the `with`  syntax ?

FredLL-Avaiga:
@mshell99 Can you try specifying async_mode="threading" as a Gui.run parameter ?

FredLL-Avaiga:
similar to [#493](https://github.com/Avaiga/taipy-gui/issues/493)

mshell99:
When I use async_mode="threading" the notify appears correctly

PR #500, opened 2022-11-24
fix gevent thread stuck (#458)

dinhlongviolin1:
Solve #458

This now works
```python
from src.taipy.gui import Gui, notify

import time

sleep_value = 1

md = """
<|{sleep_value}|number|>

<|Wait|button|on_action=wait|>
"""

def wait(state):
    notify(state,'i', 'First')
    time.sleep(state.sleep_value)
    notify(state,'i', 'Last')
    
Gui(md).run()
```


src/taipy/gui/gui.py
```diff
@@ -25,6 +25,11 @@
 from types import FrameType
 from urllib.parse import urlparse
 
+if util.find_spec("gevent"):
```
FredLL-Avaiga: What happens if async_mode is set to something else than gevent ?

src/taipy/gui/gui.py
```diff
@@ -1499,15 +1497,14 @@ def run(
...
                 Valid values are:</br>
 
                 - "threading": Use the Flask Development Server. This allows the application to use
                   the Flask reloader (the *use_reloader* option) and Debug mode (the *debug* option).
                 - "eventlet": Use eventlet server.
                 - "gevent": Use gevent server.
-                - "gevent_uwsgi": Use uwsgi server.
-
+                </br>The default value is "threading"</br>
                 If this argument is not set, Taipy uses, in that order: `"eventlet"`, `"gevent_uwsgi"`,
```
FredLL-Avaiga: this line needs to be changed too ?

src/taipy/gui/server.py
```diff
@@ -180,7 +181,20 @@ def _is_port_open(self, host, port) -> bool:
         sock.close()
         return result == 0
 
+    def _apply_patch(self):
```
FredLL-Avaiga, line 208: Nice ! 
A way to detect if patching has been done

FredLL-Avaiga (approved):

src/taipy/gui/server.py
dinhlongviolin1: Thanks!

src/taipy/gui/gui.py
dinhlongviolin1: yes imo. gevent and eventlet are what will mainly be used

src/taipy/gui/server.py
```diff
@@ -193,6 +205,18 @@ def _is_port_open(self, host, port) -> bool:
...
+        if self._get_async_mode() == "gevent" and util.find_spec("gevent"):
+            from gevent import monkey
+
+            if not monkey.is_anything_patched():
+                monkey.patch_all(ssl=False, thread=False)
+        if self._get_async_mode() == "eventlet" and util.find_spec("eventlet"):
```
FabienLelaquais, line 214: That would be an 'else', right?

FabienLelaquais (approved):
Neat!!

src/taipy/gui/server.py
dinhlongviolin1: it is definitely and 'and'. Both conditions need to be met before any furhter execution

Issue #501, opened 2022-11-25
Implement a gitignore like file restriction


FredLL-Avaiga:
**What would that feature address**
if the file .taipyignore file is present at the root level of the taipy application, files matching the conditions should not be served by the web server.

***Description of the ideal solution***
.taipyignore support the same feature as .gitignore
if the file is not present, every file in the directory tree is accessible

***Other options***
leave this to a thrid party server that would be set up "around" the taipy app

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
this overrides #498

PR #502, opened 2022-11-25
taipyignore file (#501)

FredLL-Avaiga:
example of .taipyignore file
```
**/*.py*
.*ignore

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,7 +122,6 @@ dataset
 # Demos folder
 demo*/
 !demos/
-.taipyignore
 
 # Docker local dev
 docker-compose-dev*.yml
--- a/.gitignore
+++ b/.gitignore
@@ -122,6 +122,7 @@ dataset
 # Demos folder
 demo*/
 !demos/
+.taipyignore
 
 # Docker local dev
 docker-compose-dev*.yml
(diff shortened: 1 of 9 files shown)


.gitignore
```diff
@@ -122,6 +122,7 @@ dataset
 # Demos folder
 demo*/
 !demos/
+.taipyignore
```
FabienLelaquais: Really?

FabienLelaquais (approved):
😍

.gitignore
FredLL-Avaiga: no

src/taipy/gui/server.py
```diff
@@ -74,6 +75,17 @@ def handle_message(message) -> None:
             elif "type" in message:
                 gui._manage_message(message["type"], message)
 
+    def __is_ignored(self, file_path: str) -> bool:
+        if not hasattr(self, "_ignore_matches"):
+            ignore_file = (pathlib.Path(__main__.__file__).parent / ".taipyignore") if hasattr(__main__, "__file__") else None
```
dinhlongviolin1: You might want to move `.taipyignore` filename as a constant

dinhlongviolin1 (approved):

src/taipy/gui/server.py
FredLL-Avaiga: a variable at least, you're right

Issue #511, opened 2022-12-01
BUG-Array of Array 


FredLL-Avaiga:
**Description**
when passing an array of array (same length) as data for table or chart, the data is rotated.
it shouldn't be :-)

**How to reproduce**
A code fragment
```
from taipy import *;
md = "<|{[1, 3, 5], [2, 4, 6]}|table|>"
...
```

Should show
|0|1|2|
|---|---|---|
|1|3|5|
|2|4|6|

**Runtime environment**
Please specify relevant indications.
 - OS: [e.g. Linux, Windows] and version
 - Browser: [e.g. Chrome, Edge, Safari] and version
and any other relevant information.

PR #512, opened 2022-12-01
#511 do not rotate data of array of array

FredLL-Avaiga:
and #513

### DIFF
--- a/src/taipy/gui/data/array_dict_data_accessor.py
+++ b/src/taipy/gui/data/array_dict_data_accessor.py
@@ -35,7 +35,7 @@ class _ArrayDictDataAccessor(_PandasDataAccessor):
             if len(types) == 1 and next(iter(types), None) == list:
                 lengths = {len(x) for x in value}
                 return (
-                    pd.DataFrame({str(i): v for i, v in enumerate(value)})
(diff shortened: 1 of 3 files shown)


src/taipy/gui/data/array_dict_data_accessor.py
```diff
@@ -29,13 +29,13 @@ def get_supported_classes() -> t.List[str]:
 
     def __get_dataframe(self, value: t.Any) -> t.Union[t.List[pd.DataFrame], pd.DataFrame]:
         if isinstance(value, list):
-            if isinstance(value[0], (str, int, float, bool)):
+            if not value or isinstance(value[0], (str, int, float, bool)):
                 return pd.DataFrame({"0": value})
```
FabienLelaquais, line 33: Is this ok when value is None?

FabienLelaquais (approved):
👍

dinhlongviolin1 (approved):

src/taipy/gui/data/array_dict_data_accessor.py
FredLL-Avaiga: value is not None as it is a List :-)

PR #514, opened 2022-12-02
add 3.11 tests

dinhlongviolin1:


.github/workflows/publish.yml
```diff
@@ -10,7 +10,7 @@ on:
 jobs:
   test-package:
     timeout-minutes: 20
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-20.04
```
FredLL-Avaiga: Why going back to ubuntu 20 ? LTS is 22 at the moment

Pipfile
```diff
@@ -43,7 +43,7 @@ pytest-cov = "*"
 pytest-mock = "3.6.1"
 tox = "==3.25.1"
 twine = "*"
-playwright = "<1.28"
```
FredLL-Avaiga, line 46: playwright over 1.28 is pulling a more recent version of gevent 
But I suppose that we're using the latest gevent now ?

setup.py
```diff
@@ -53,7 +53,7 @@
         "python-magic>=0.4.24,<0.5;platform_system!='Windows'",
         "python-magic-bin>=0.4.14,<0.5;platform_system=='Windows'",
     ],
-    "arrow": ["pyarrow>=9.0,<10.0"],
+    "arrow": ["pyarrow>=10.0.1,<11.0"],
```
FredLL-Avaiga, line 56: Should we update arrow to 10 on the front end side too then ?

FredLL-Avaiga (changes requested):
ubuntu 22 is LTS

.github/workflows/publish.yml
dinhlongviolin1: yes I know. I am just testing to see why it is not passing the tests on github actions but my computer is passing

Pipfile
dinhlongviolin1: we are using the latest now so I think the restriction can be lifted

setup.py
dinhlongviolin1: I will do it!

src/taipy/gui/gui.py
```diff
@@ -734,10 +734,10 @@ def __send_ws(self, payload: dict) -> None:
                     payload,
                     to=self.__get_ws_receiver(),
                 )
-                time.sleep(0)
+                time.sleep(0.001)
```
FredLL-Avaiga, line 737: :-o

FredLL-Avaiga (approved):

src/taipy/gui/gui.py
dinhlongviolin1: 0.001 is the default. If I dont put it then a warning will be shown in gevent 22 :D

FabienLelaquais (approved):
👌

Issue #522, opened 2022-12-05
:sparkles: Have the ability to select rows in table


gmarabout:
**What would that feature address**
It would be useful to be able to select rows in a Taipy table, even if the table is not editable.

***Description of the ideal solution***
* Have a "selection_mode" property with possible values: None (the default), single_row, multiple_rows.
* Have a callback function to be called on "selection changed" event
* Should be linked to the same property value as the existing "selected" property

***Caveats***

***Other options***
* one can add a boolean field on row object, but:
* can be editable only if the whole table is editable (which is not always what we want)
* is currently shown as a toggle (would expect a check box)

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
To allow edition of only one column, you can specify `|not editable|editable[boolean column name]|`
But I'll add an on_action property (mono selection for now)

gmarabout:
Thanks, but can you give an example of what is expected for `|editable[boolean column name]|` ?

FredLL-Avaiga:
some properties can be indexed by column name.
editable is one of them:
- editable without [...] applies to the whole table
- editable with [a column name] applies only the specified column

PR #523, opened 2022-12-05
#522 single row selection

FredLL-Avaiga:
- via on_action
- force columns dict keys to be str
- fix rowIndex parameter for on_delete and on_action from page based to *not* paged base


FabienLelaquais (approved):
Neat!

dinhlongviolin1 (approved):

Issue #526, opened 2022-12-07
BUG-page_size property is not working


FredLL-Avaiga:
**Description**
page_size=2 =>
- page_size in not impacted
- page size options is not impacted
- page size is page size +1

**How to reproduce**
A code fragment
```
...
<|{data}|table|page_size=2|>
...
```

**Expected behavior**
page size is respected

**Runtime environment**
Please specify relevant indications.
 - OS: [e.g. Linux, Windows] and version
 - Browser: [e.g. Chrome, Edge, Safari] and version
and any other relevant information.

Issue #525, opened 2022-12-06
:bug: Issue when using the "editable[column name]" property with empty array data


gmarabout:
**Description**
We can make a single table column editable using the following property: `editable[column]=True`.
However this causes a stack when there is no data.

**How to reproduce**
Run the following code:
```
from taipy.gui import Gui

data = []
columns = {"0": {"title": "Name"}, "1": {"title": "Activated"}}


page = """
<|{data}|table|columns={columns}|editable[2]=True|>
"""

Gui(page=page).run(port=8000)
```

You get:
<img width="1060" alt="Screenshot 2022-12-06 at 16 09 30" src="https://user-images.githubusercontent.com/26060559/205949155-439c9133-5f04-4b2d-9ff3-858ab87f153b.png">

And in the console:

```
 * Server starting on http://127.0.0.1:8000
[2022-12-06 16:11:06,859][Taipy][ERROR] Exception on /taipy-jsx/TaiPy_root_page [GET]
Traceback (most recent call last):
  File "/Users/gmarabout/Library/Caches/pypoetry/virtualenvs/taipy-demo-job-monitoring-PuBFsu6N-py3.10/lib/python3.10/site-packages/flask/app.py", line 2525, in wsgi_app
    response = self.full_dispatch_request()
  File "/Users/gmarabout/Library/Caches/pypoetry/virtualenvs/taipy-demo-job-monitoring-PuBFsu6N-py3.10/lib/python3.10/site-packages/flask/app.py", line 1822, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/Users/gmarabout/Library/Caches/pypoetry/virtualenvs/taipy-demo-job-monitoring-PuBFsu6N-py3.10/lib/python3.10/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/Users/gmarabout/Library/Caches/pypoetry/virtualenvs/taipy-demo-job-monitoring-PuBFsu6N-py3.10/lib/python3.10/site-packages/flask/app.py", line 1820, in full_dispatch_request
    rv = self.dispatch_request()
  File "/Users/gmarabout/Library/Caches/pypoetry/virtualenvs/taipy-demo-job-monitoring-PuBFsu6N-py3.10/lib/python3.10/site-packages/flask/app.py", line 1796, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
[...]

PR #527, opened 2022-12-07
Paginated table properties (#525 #526)

FredLL-Avaiga:
- #525 indexed property not found 
- #526 page_size property
- force columns dict key as non str


FabienLelaquais (approved):
Looks good to me!

dinhlongviolin1 (approved):
Interesting one!

Issue #531, opened 2022-12-08
BUG-Table: rowIndex is not valid when sorting


FredLL-Avaiga:
**Description**
the rowIndex parameter passed to on_action, on_edit or on_delete callbacks is not valid when the table is sorted.

**How to reproduce**
A code fragment
```
from taipy.gui import Gui, State

data = [["Greg", "Montpellier", "France"], ["Greg", "Montpellier", "France"], ["Bill", "Seattle", "USA"], ["Bill", "Seattle", "USA"]]
columns = {0: {"title": "Name"}, 1: {"title": "City"}, 2: {"title": "Country"}}
def on_selection(state: State, var_name: str, action: str, payload: dict):
    print(payload)

page = """
##Table to sort

<|{data}|table|columns={columns}|on_action=on_selection|page_size=2|>
"""

Gui(page=page).run(port=8000)
```

- Sort by column so that the order of the rows changes
- click on a row and check the rowIndex in the python console
...

**Expected behavior**
rowIndex should allow to access the row from original data

Issue #530, opened 2022-12-08
:bug: BUG-Cannot sort table using Python lists


gmarabout:
**Description**
Using a table bound to a Python list for its data, the sort would not work.

**How to reproduce**
```
from taipy.gui import Gui

data = [[f"Name {i}", i % 3 == 0] for i in range(1000)]
columns = {"0": {"title": "Name"}, "1": {"title": "Activated"}}


page = """
<|{data}|table|columns={columns}|>
"""

Gui(page=page).run(port=8000)
```
 - Execute the script here about (using the latest `develop` version)
 - Try sort a column by clicking on its header
 Log in the console:
```
UserWarning: Decoding Message has failed: {'type': 'DU', 'name': '_TpD_tpec_TpExPr_data_TPMDL_1', 'payload': {'start': 0, 'end': 100, 'orderby': '1', 'sort': 'desc', 'aggregates': [], 'styles': {}, 'handlenan': False, 'filters': [], 'columns': ['0', '1'], 'pagekey': '0-100-0,1-1-desc'}, 'propagate': True, 'client_id': '20221128162535407204-0.7325782016898321', 'ack_id': 'e2dbe19c-37e0-4136-8e14-c370b76819e7', 'module_context': '__main__'}
'1'
```
**Expected behavior**
* The sort works
* No error in the console

**Runtime environment**
Please specify relevant indications.
 - OS: MacOS
 - Browser: Chrome
 - Python 3.10.5
 - Taipy 2.1.0.dev0

FredLL-Avaiga:
link to the columns not having name thus we use index.

Issue #528, opened 2022-12-08
:bug: BUG-Using the `style` table property causes the table to stop displaying


gmarabout:
**Description**
When using the `style` property as described (here)[https://docs.taipy.io/en/latest/manuals/gui/viselements/table/#show-a-table], the table stops displaying and the console displays a Web socket warning:  

```
UserWarning: Exception raised in Web Socket communication in '<module>':
'<' not supported between instances of 'str' and 'int'
```

**How to reproduce**

```
from taipy.gui import Gui

data = [["Item1", True], ["Item2", True], ["Item3", False]]
columns = {"0": {"title": "Name"}, "1": {"title": "Activated"}}

page = """
<|{data}|table|style={lambda state, idx, row: "red-row" if idx % 2 == 0 else "blue-row"}|columns={columns}|>
"""

Gui(page=page).run(port=8000)
```
Run the above code.
  - Notice that the presence of an style sheet does not change the behaviour

The view is the following: 
<img width="1524" alt="Screenshot 2022-12-08 at 10 24 15" src="https://user-images.githubusercontent.com/26060559/206408780-c198b5a1-911f-4768-892a-cb1114b135c5.png">

The console:
```
UserWarning: Exception raised in Web Socket communication in '<module>':
'<' not supported between instances of 'str' and 'int'
```

...

**Expected behavior**
* The style is applied
* The table displays properly

**Runtime environment**
 - OS: MacOS
 - Browser: Chrome
 - Python 3.10.5

FredLL-Avaiga:
this happens when columns don't have name (index is used) 
And we need to add a new column (which has a name)
The json encoder tries to sort the keys of the row object which is not necessary and time consuming.

PR #532, opened 2022-12-08
column with no name and rowIndex when sorted

FredLL-Avaiga:
- #528 
- #530 
- #531


FabienLelaquais (approved):
Bad enough to deserve a test don't you think?

dinhlongviolin1 (approved):

PR #533, opened 2022-12-12
refactor + config resolver

dinhlongviolin1:


FredLL-Avaiga (approved):
quite a bit of shuffling around

FabienLelaquais (approved):
I like what I see and it seems to cover my fears.
Well done!

Issue #535, opened 2022-12-12
Chart: Set a default value for template[dark] in config


FredLL-Avaiga:
**Description**
Would make life easier for users that want to use Dark theme
https://github.com/Avaiga/taipy-gui/issues/505

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #536, opened 2022-12-12
chart dark mode OOTB (#535)

FredLL-Avaiga:
nothing to do by default
- default value in front-end
- config value as default


dinhlongviolin1 (approved):

drake-stamos (approved):
LGTM!

Issue #538, opened 2022-12-13
BUG-Wrong display of bar trace


FlorianJacta:
**Description**
Bars of the bar trace begins at the wrong position (on top of the line trace).

**How to reproduce**
```
from taipy import Gui 
import pandas as pd

x = [i for i in range(10)]
y1 = [i*0.5 for i in range(10)]
y2 = [i*0.5 for i in range(10)]
y3 = [i*1.5 for i in range(10)]

data = pd.DataFrame({"x":x,"y1":y1,"y2":y2,"y3":y3})

Gui('<|{data}|chart|x=x|y[1]=y1|y[2]=y2|y[3]=y3|type[1]=bar|>').run()
```

**Expected behavior**
The bar chart should begin at 0 and not on top of the line trace value.

![newplot(1)](https://user-images.githubusercontent.com/98709993/207293362-222ead35-3a76-4e32-92e6-cfac24ddacf9.png)

**Runtime environment**
Taipy version: taipy-gui-2.1.0.dev0

PR #539, opened 2022-12-15
less default cols (#538)

FredLL-Avaiga:
- set default cols only if no col defined for the trace


dinhlongviolin1 (approved):
Nicely done

FabienLelaquais (approved):
👌

PR #542, opened 2022-12-18
New chart landing page

haotran-avaiga:
I have some issues when working on this ticket:
- we just have the document for some charts.
- I don't know how to put the path of each chart in a tag
- I have to use HTML to style the list of charts, I am not sure it is the best way
- I know we generate the table of properties, but I don't know how to put it into the MD file.

I appreciate it if you put your comments here so I can fix them.


FredLL-Avaiga (commented):
any dark them images Or am I mistaken ?

haotran-avaiga:
> any dark them images Or am I mistaken ?

sorry no, all images are taken from the design file, so there is no dark theme in image. Fabien told me we will use Taipy to generate the images (both light and dark versions)

gui/doc/chart.md
```diff
@@ -3,30 +3,268 @@ Displays data sets in a chart or a group of charts.
 The chart control is based on the [plotly.js](https://plotly.com/javascript/)
 graphs library.
 
-A chart control can hold several traces, that can display individual data sets.
+Plotly is an amazing graphing library that provides a very large number of visual
```
FredLL-Avaiga: Amazing ?

gui/doc/chart.md
```diff
@@ -3,30 +3,268 @@ Displays data sets in a chart or a group of charts.
...
+    value1,
+    value2,
+    value3
+]
+    
+page = "<|...|chart|...|index_property={values}|...|>"
+```
+is equivalent to
+```
+page = "<|...|chart|...|index_property[1]={value1}|index_property[2]={value2}|index_property[3]={value3}|...|>"
```
FredLL-Avaiga, line 234: <|...|chart|...|index_property={value1}|index_property[1]={value1}|index_property[2]={value2}|index_property[3]={value3}|...|>

FredLL-Avaiga (approved):

gui/doc/chart.md
FabienLelaquais: I'm sorry. I don't get your point...
(except that 'index_property' should be 'indexed_property' everywhere, which is too long)

gui/doc/chart.md
FabienLelaquais: Hum. A bit too much?
(that's what I think, nevertheless)

FabienLelaquais: Would  'impressive' do? 'Comprehensive'?

FredLL-Avaiga: Comprehensive I like

gui/doc/chart.md
FredLL-Avaiga: when you specify an array, the first value of the array becomes the default value.
ie if your array is shorter than the number of traces, the extra traces inherit from the 1st value of the array (like they would if the non index property was not an array)

gui/doc/chart.md
FabienLelaquais: 'vast' it is...

gui/doc/chart.md
FabienLelaquais: Yes.
So I guess your
`index_property={value1}|index_property[1]={value1}`
was meant to be
`index_property={value1}`

FredLL-Avaiga: Which is the same in the end (and might be clearer ?)

PR #553, opened 2022-12-30
Fix - Misuse of NotImplemented

trgiangdo:
Fix similar issue on https://github.com/Avaiga/taipy-core/issues/370


jrobinAV (approved):
Good for me

PR #570, opened 2023-01-10
Align chart sub-types no matter their names

FabienLelaquais:
CSS changes from Jaufrey


ghost (approved):
Seems like a good piece of work 👍

FredLL-Avaiga (approved):
Less is more

Issue #574, opened 2023-01-16
BUG-Format not applied for text control


FlorianJacta:
**Description**
The format is not applied to the number for the _text_ control.

**How to reproduce**
```
from taipy import Gui 
import numpy as np
import pandas as pd

data = pd.DataFrame({'values':[0,0,1]})

Gui("<|{np.mean(data['values'])}|text|format=%.2f|>").run()
```

**Expected behavior**
We should get 0.33 for the displayed number and not 0.3333333...

**Runtime environment**
Please specify relevant indications.
 - Taipy dev <2.1

aazimh:
Just came here searching for something similar. This is what I have:

`<|{total_pred}|text|class_name=test|format=%,.0f|>`

The value of `total_pred` is `8075105.59837587` and it displays [like this](https://imgur.com/BzBEnqc).

i.e. it rounds to 3 decimal places when I need it to be none. I might be doing something wrong in terms of the format string but open to comments. 

I did notice that it works as it should when the raw number in question does not have a decimal component (i.e. is `45252.0` for example). If the decimal becomes non-zero, then it doesn't work.

I'm on Taipy 2.0.0.

aazimh:
I did a bit more fiddling and now it appears to work the other way (not sure what I changed specifically, but I now cast all my values as floats). 

It seems to be linked to the type of value you pass to the text element - if it's an integer and you apply a number format, it doesn't seem to work properly, but if you cast the same value as a float and apply a number format then it works fine. To be fair the documentation mentions floating point values specifically, but I just expected there to be some exception handling that would work with integers. 

This may not be related to the initial issue (since that is not an integer) so apologies if I hijacked it, just hoping my experience is useful in debugging.

FredLL-Avaiga:
Indeed <|{float(np.mean(data['values']))}|text|format=%.2f|> gives the expected result

FredLL-Avaiga:
Hi @Aazimh,
the package used to format the value in the browser is [sprinf-js](https://www.npmjs.com/package/sprintf-js).
The format %,.0f is not valid for sprintf-js but %.0f works (I beleive you should have a warning in the browser console).
The datatype is calculated at render from the value but int or float should be fine.
The original question shows that we do not support (yet ?) numpy.float64
A quick workaround would be (as you discovered ?) to cast the result to a python float

aazimh:
[...]

PR #575, opened 2023-01-18
#574 manage numpy and other types

FredLL-Avaiga:
Use pandas to get the types


dinhlongviolin1 (approved):
Concise

FabienLelaquais (approved):
Looks neat.
I suspect this new import doesn't hurt hu?

Issue #578, opened 2023-01-23
BUG-Selection in charts


FabienLelaquais:
**Description**

Only single-point selection seems to work in charts.
Lasso and box selections are not handled properly.

**How to reproduce**

Run this code:

```
from taipy import Gui
import random

x = list(range(0, 21))
y = [random.uniform(1, 20) for _ in x]
data = {
    "x": x,
    "y": y
}
selected_indices = []

def on_change(state, var, val):
    print(f"Var={var} Val={val}")

Gui("<|{data}|chart|selected={selected_indices}|>").run(run_browser=False)
```

Selecting a single data point triggers the `on_change()` function alright.

However, box and lasso selections, although point are properly selected in the chart, do not trigger the callback.

**Expected behavior**

The selection array variable should be updated according to the actual selection.

PR #580, opened 2023-01-24
Chart: Selection (#578)

FredLL-Avaiga:
Addresses #578.

Not every mean of selection modifies the *selected* property.

PS: Deselection works only when there's one trace as the plotly deselect event doesn't give any context.


FabienLelaquais (approved):
Nice spot!

Issue #579, opened 2023-01-23
BUG-Charts on_range_changed called in different situations


FabienLelaquais:
**Description**

The callback function set to the *on_range_change* property of the chart control gets called in many different situations.

The *payload* parameter gets updated depending on the action:

- Pressing the Zoom In or Zoom out buttons, or requesting a box zoom properly calls the callback and set the *xaxis.range[0]*...*yaxis.range[1]* of *payload* to the appropriate new values of the axis ranges;
- Pressing the Reset Axes button also calls the callback, and sets the *xaxis/yaxis.autorange* and 'xaxis/yaxis.showspikes* properties in *payload*;
- Selecting the Zoom, Pan, Box select or Lasso select tools do invoke the callback, with the *dragmode* property of *payload* set to whatever the selected tool is;
- A lasso selection invokes the callback sets the *selections* property of *payload* to a dictionary that describes the shape of the hand-drawn lasso shape;
- may be other cases.

I think this is rather unclear: how to simply detect that we indeed are in a 'range change' situation?

**How to reproduce**

Run the following code:

```
from taipy import Gui
import random

x = list(range(0, 21))
y = [random.uniform(1, 20) for _ in x]
data = {
    "x": x,
    "y": y
}

def range_changed(state, _1, _2, payload):
    print(f"range_changed: {payload}")

Gui("<|{data}|chart|on_range_change=range_changed|>").run(run_browser=False)
```

And manipulate the chart, including the buttons in the chart menu.

**Expected behavior**

The callback should be called only when the range actually changes.
Or its name is very wrong. And the doc... inexistant.

FredLL-Avaiga:
on_range_change is triggered by the onRelayout plotly event ...
Do you think we should filter the events and not callback on some scenarios @FabienLelaquais ?

FabienLelaquais:
> on_range_change is triggered by the onRelayout plotly event ... Do you think we should filter the events and not callback on some scenarios @FabienLelaquais ?

Because of the naming of the callback, I would say yes.
Then we'll have to add (in the future) another callback or reuse on_change with documented context.

For the time being my problem is I don't know what to document, really.

FredLL-Avaiga:
I find it a bit difficult to limit this. Especially when we have a reference to the relayout event in the doc ...

PR #584, opened 2023-01-25
#579 limit on_range_change calls

FredLL-Avaiga:
call on range change only when xaxis.* is present in the relayout event

### DIFF
--- a/gui/src/components/Taipy/Chart.tsx
+++ b/gui/src/components/Taipy/Chart.tsx
@@ -413,29 +413,31 @@ const Chart = (props: ChartProp) => {
 
     const onRelayout = useCallback(
         (eventData: PlotRelayoutEvent) => {
-            onRangeChange && dispatch(createSendActionNameAction(id, { action: onRangeChange, ...eventData }));
-            if (config.decimators && !config.types.includes("scatter3d")) {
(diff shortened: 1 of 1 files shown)


gui/src/components/Taipy/Chart.tsx
```diff
@@ -413,29 +413,31 @@ const Chart = (props: ChartProp) => {
...
-                        backCols,
-                        dataKey.current,
-                        getDecimatorsPayload(
-                            config.decimators,
-                            plotRef.current,
-                            config.modes,
-                            config.columns,
-                            config.traces,
-                            eventData
+            if (Object.keys(eventData).some(k => k.startsWith("xaxis."))) {
```
FabienLelaquais, line 416: What about "|| ...yaxis...?

FabienLelaquais (approved):
👌

dinhlongviolin1 (approved):

gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: I tested the different use cases, and I could find a case when yaxis is present and not xaxis ...

FabienLelaquais: Ok then... fingers crossed

PR #590, opened 2023-01-30
bug/#127-ensure-all-init-files-are-the-same

jrobinAV:
Make sure the init file of all our repositories are the same!

### DIFF
--- a/src/taipy/__init__.py
+++ b/src/taipy/__init__.py
@@ -11,7 +11,7 @@
 
 from importlib.util import find_spec
 
-if find_spec("taipy"):
+if find_spec('taipy'):
     if find_spec("taipy.config"):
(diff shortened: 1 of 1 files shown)


src/taipy/__init__.py
```diff
@@ -11,7 +11,7 @@
 
 from importlib.util import find_spec
 
-if find_spec("taipy"):
+if find_spec('taipy'):
```
FabienLelaquais: The convention everywhere is to use " for strings...

FabienLelaquais (approved):
But conventions are not preserved...

src/taipy/__init__.py
FabienLelaquais: Done in second commit.

Issue #588, opened 2023-01-26
BUG- on_navigate() switches to the requested page content, but the URL is not updated in the browser


Dr-Irv:
**Description**
If you use `on_navigate()` to change the page, the URL in the browser is not updated to the correct place.

**How to reproduce**
```python
from taipy import Gui
from taipy.gui import State, notify

allow_page2 = False
last_request = "Unknown"


def on_navigate(state: State, page_name: str) -> str:
    state.last_request = page_name
    if page_name == "page2" and not allow_page2:
        notify(state, "error", "cannot go to page2 before page1")
        return "page1"
    return page_name


page1_md = """## Page 1
Last requested page is <|{last_request}|text|>
"""
page2_md = "## Page 2"
navbar = "<|navbar|lov={[('/page1', 'First Page'), ('/page2', 'Second Page')]}|" ">"


if __name__ == "__main__":
    gui = Gui()

    gui.add_pages({"/": navbar, "page1": page1_md, "page2": page2_md})

    gui.run(
        page=navbar,
        async_mode="threading",
        title="Notify Demo",
        debug=True,
        use_reloader=True,
    )

```

When you start the app, the URL shown in the browser is `http://127.0.0.1:5000/page1`.  When you click on `page2`, it should switch to show that URL.   The content does switch, but the URL doesn't change.  Also, the navbar doesn't update either (but that's probably because the URL is wrong).

**Expected behavior**
The URL in the browser should change to the requested page.

**Screenshots**
At beginning of app, it looks like this:
![image](https://user-images.githubusercontent.com/15113894/214907010-712e2c59-91e3-4281-ac95-b57b80e4a101.png)

If you click on "page2", you get this.  Note that the content is correct, but the URL at the top is wrong.
![image](https://user-images.githubusercontent.com/15113894/214907130-96c90005-a37b-47a4-848f-fcf99f7b46e7.png)

**Runtime environment**
Please specify relevant indications.
 - OS:  Windows 10
 - Browser: Chrome

Dr-Irv:
It seems that if I call `navigate()` in the `on_navigate()` function, it works, i.e.,
```python
def on_navigate(state: State, page_name: str) -> str:
    state.last_request = page_name
    if page_name == "page2" and not allow_page2:
        notify(state, "error", "cannot go to page2 before page1")
        navigate(state, "page1")
        return "page1"
    return page_name
```
[...]

Issue #585, opened 2023-01-25
BUG- `navbar` control doesn't call on_change method, and docs are misleading about how to define the LOV


Dr-Irv:
**Description**
If you want to use the `navbar` control, there are 2 issues:
1. If you specify an `on_change()` function, it never gets called.
2. If you want to navigate to the pages in your application, then the first element of the LOV needs to be preceded with a `"/"`

**How to reproduce**
A code fragment
```python
from taipy import Gui
from taipy.gui import State

from typing import Any


def navbar_change(s: State, var_name: str, value: Any):
    print("navbar changed")


page1_md = "## Page 1"
page2_md = "## Page 2"
navbar = (
    "<|navbar|lov={[('/page1', 'First Page'), ('/page2', 'Second Page')]}|"
    + "on_change={lambda s,vn, va: navbar_change(s, vn, va)}|"
    ">"
)


if __name__ == "__main__":
    gui = Gui()

    gui.add_pages({"/": navbar, "page1": page1_md, "page2": page2_md})

    gui.run(
        page=navbar,
        async_mode="threading",
        title="Notify Demo",
        debug=True,
        use_reloader=True,
    )

```

**Expected behavior**

The `on_change` function should be called.  The docs should say that if you want your page to be called, then the LOV should be preceded with a "/".  In my example above, if you use `lov={[('page1', 'First Page'), ('page2', 'Second Page')]}` then it always opens up the new page in a second window, and the selection isn't correctly reflected in the navbar.

**Screenshots**
N/A

**Runtime environment**
Please specify relevant indications.
 - OS: Windows 10
 - Browser:Chrome

FlorianJacta:
Thank you. I guess the problem on `/` for the navbar is either just a doc issue or an issue on the visual element.

FredLL-Avaiga:
I think the leading / has been addressed in the doc
on_change is not supported on navbar and the doc should be fixed about it

FredLL-Avaiga:
Doc change

Dr-Irv:
As a related issue, it would be nice if `navbar` could be used to "navigate" to a `part` with a partial.  I'd like to have a `navbar` at the top of the page that navigates to a page, but then within that page, have a `navbar` that changes a `partial` within the page.

If I should open up a separate issue for this, let me know.

FredLL-Avaiga:
You can create another issue.
But Navbar is really dedicated to navigating to pages.
I guess we could find another component that does what you need

[...]

Issue #581, opened 2023-01-24
BUG- can't dismiss a notification with `taipy.gui.notify(state, "")`


Dr-Irv:
**Description**
The docs indicate that passing a notification type of `""` will dismiss the last notification, but that doesn't seem to work.

**How to reproduce**

```python
from taipy import Gui
from taipy.gui import State, notify

result = "UNKNOWN"
notified = False
show_result = False


def checker(state: State):
    if state.result == "Bad":
        notify(state, "error", "You cannot enter 'Bad'", duration=10000)
        return
    else:
        notify(state, "")  # should remove the notification
    state.show_result = True


page = """<|{result}|input|on_action={lambda s: checker(s)}|>

<|part|render={show_result}|

<|Result is|text|>
<|{result}|text|>

|>
"""

if __name__ == "__main__":
    Gui(page=page).run(
        async_mode="threading",
        title="Notify Demo",
        debug=True,
        use_reloader=True,
    )
```

Type in the word `Bad` into the box, and you will get a notification. Then type anything else, and you will see two notifications, the red one and a blue one.

**Expected behavior**
The red notification should disappear.

**Screenshots**

![image](https://user-images.githubusercontent.com/15113894/214411537-eed93b81-65e3-47cf-a94c-8a7f8b4beaf8.png)
**Runtime environment**
Please specify relevant indications.
 - OS: Windows 10
 - Browser: Chrome 109.0.5414.74

FlorianJacta:
Thank you! Indeed, the notification behavior does not reflect what is written in the documentation for the _notify_ function. We will take a look at it.

PR #591, opened 2023-01-30
#581 dismiss notification

FredLL-Avaiga:
#588 on_navigate change URL
#585 navbar doc

### DIFF
--- a/gui/src/components/pages/TaipyRendered.tsx
+++ b/gui/src/components/pages/TaipyRendered.tsx
@@ -87,7 +87,7 @@ const TaipyRendered = (props: TaipyRenderedProps) => {
                 }
                 dispatch(createModuleContextAction(result.data.context));
             })
-            .catch((error) => setJSX(`<h1>${error.response?.data || `No data fetched from backend from ${path === "/TaiPy_root_page" ? "/" : path}`}</h1><h2></h2>${error}`));
+            .catch((error) => setJSX(`<h1>${error.response?.data || `No data fetched from backend from ${path === "/TaiPy_root_page" ? "/" : path}`}</h1><br></br>${error}`));
         }
     }, [path, state.id, dispatch, partial, fromBlock]);
 
--- a/gui/src/components/pages/TaipyRendered.tsx
+++ b/gui/src/components/pages/TaipyRendered.tsx
@@ -87,7 +87,7 @@ const TaipyRendered = (props: TaipyRenderedProps) => {
                 }
                 dispatch(createModuleContextAction(result.data.context));
             })
-            .catch((error) => setJSX(`<h1>No data fetched from backend from ${path}</h1><br></br>${error}`));
+            .catch((error) => setJSX(`<h1>${error.response?.data || `No data fetched from backend from ${path === "/TaiPy_root_page" ? "/" : path}`}</h1><h2></h2>${error}`));
         }
     }, [path, state.id, dispatch, partial, fromBlock]);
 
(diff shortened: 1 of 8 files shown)


gui/src/components/pages/TaipyRendered.tsx
```diff
@@ -87,7 +87,7 @@ const TaipyRendered = (props: TaipyRenderedProps) => {
                 }
                 dispatch(createModuleContextAction(result.data.context));
             })
-            .catch((error) => setJSX(`<h1>No data fetched from backend from ${path}</h1><br></br>${error}`));
+            .catch((error) => setJSX(`<h1>${error.response?.data || `No data fetched from backend from ${path === "/TaiPy_root_page" ? "/" : path}`}</h1><h2></h2>${error}`));
```
FabienLelaquais: What the use of "&lt;h2&gt;&lt;/h2&gt;"  ?

gui/src/context/taipyReducers.ts
```diff
@@ -673,18 +673,19 @@ export const createTimeZoneAction = (timeZone: string, fromBackend = false): Tai
...
             return "error";
         case "w":
             return "warning";
         case "s":
             return "success";
-        default:
+        case "i":
             return "info";
     }
+    return aType;
```
FabienLelaquais, line 690: So the user can provide his own alert type?

dinhlongviolin1 (approved):
LGTM

src/taipy/gui/gui.py
```diff
@@ -1369,8 +1370,12 @@ def __render_page(self, page_name: str) -> t.Any:
         if hasattr(self, "on_navigate") and callable(self.on_navigate):
             try:
                 nav_page = self.on_navigate(self.__get_state(), page_name)
-                if not isinstance(nav_page, str):
-                    warnings.warn(f"on_navigate() returned a invalid page name '{nav_page}'.")
+                if nav_page != page_name:
+                    if isinstance(nav_page, str):
+                        if self._navigate(nav_page):
+                            return ("Root page cannot be re-routed by on_navigate", 302)
```
FabienLelaquais: "on_navigate" -> "on_navigate()."

What does this mean? re-routed??
You cannot navigate from '/' to another page?
So that would be "Cannot navigate from root page in on_navigate()."

src/taipy/gui/gui.py
```diff
@@ -1369,8 +1370,12 @@ def __render_page(self, page_name: str) -> t.Any:
...
+                    else:
+                        warnings.warn(f"on_navigate() returned a invalid page name '{nav_page}'.")
```
FabienLelaquais: "a invalid" -> "an invalid"

FabienLelaquais (approved):
Looks good!
A question though.

gui/src/components/pages/TaipyRendered.tsx
FredLL-Avaiga: make vertical space: will change

gui/src/context/taipyReducers.ts
FredLL-Avaiga: no, I'll change this

src/taipy/gui/gui.py
FredLL-Avaiga: we cannot replace the destination page in on_navigate when the root page is requested

FredLL-Avaiga: This message is only visible when the root page is requested

Issue #529, opened 2022-12-08
BUG-Display bug when a table is lazily populated


gmarabout:
**Description**
If a table starts empty, and then populated after a callbacl/action by modifying the state,
the table may not display properly.
Refresing the page will then display the table well.

**How to reproduce**

```
from taipy.gui import Gui

data = []
columns = {"0": {"title": "Name"}, "1": {"title": "Activated"}}


def load_data(state):
    print("Clicked!")
    state.data = [[f"name {i}", True] for i in range(1000)]


page = """
<|{data}|table|columns={columns}|>
<|Load Data|button|on_action=load_data|>
"""

Gui(page=page).run(port=8000)
```

- Run the script
- The table stats "empty" since the `data` is an empty array
- Click the button which calls `load_data`
- See the table display
- Refresh the page for a good display

Initial state (OK):
<img width="1524" alt="Screenshot 2022-12-08 at 10 36 31" src="https://user-images.githubusercontent.com/26060559/206411406-e6122898-08ca-4728-93fe-c8ef48477f12.png">

After clicking the button (Not OK):
<img width="1524" alt="Screenshot 2022-12-08 at 10 36 08" src="https://user-images.githubusercontent.com/26060559/206411498-5bfa430c-bc9a-48be-9a8b-0a017dc3dfa5.png">

After refreshing the page (now OK):
<img width="1524" alt="Screenshot 2022-12-08 at 10 36 13" src="https://user-images.githubusercontent.com/26060559/206411549-2bfb0a1e-eb1e-47da-b2fb-38e91305e8af.png">

**Expected behavior**
The table should always display properly

**Runtime environment**
Please specify relevant indications.
 - OS: MacOS
 - Browser: Chrome
 - Python 3.10.5

FredLL-Avaiga:
Not sure we'll fix this one as the columns definition is static and calculated at the first render.

gmarabout:
Hello @FredLL-Avaiga 👋🏻 
Thank you for your feedback. 
However, I don't think we should rely on the fact that a table would always be initially populated. The use-case where the table starts "empty" does not look that exotic to me...

FredLL-Avaiga:
you could have an empty dataframe/dict with defined columns

PR #614, opened 2023-02-20
#529 allow for dynamic definition of columns

FredLL-Avaiga:
via reset property (@FabienLelaquais do you have/want a better name ?)

Example: switch column order
```
from taipy.gui import Gui

data = {"Col 1": [f"name {i}" for i in range(1000)], "Col 2": [True if i % 2 else False for i in range(1000)]}

columns="Col 1;Col 2"

def switch_cols(state):
    state.columns = "Col 2;Col 1" if state.columns == "Col 1;Col 2" else "Col 1;Col 2"

page = """
<|{data}|table|reset|columns={columns}|>
<|Switch Cols|button|on_action=switch_cols|>
"""

Gui(page=page).run()
```

### DIFF
--- a/src/taipy/gui/renderers/builder.py
+++ b/src/taipy/gui/renderers/builder.py
@@ -464,7 +464,7 @@ class _Builder:
             else:
                 warnings.warn(f"{self.__element_name} {name}[{k}] is not in the list of displayed columns")
 
-    def _get_dataframe_attributes(self, date_format="MM/dd/yyyy", number_format=None) -> '_Builder':
+    def _get_dataframe_attributes(self, date_format="MM/dd/yyyy", number_format=None) -> '_Builder': # noqa: C901
         columns, col_types, reset_fn_hash = self.__gui._calculate_table_columns(_is_boolean_true(t.cast(bool, self.__attributes.get("reset", False))), self.__hashes.get("reset"),
             self.__attributes.get("data"), self.__hashes.get("data", ""),
             self.__attributes.get("columns", {}), self.__hashes.get("columns"),
--- a/src/taipy/gui/renderers/builder.py
+++ b/src/taipy/gui/renderers/builder.py
@@ -465,12 +465,12 @@ class _Builder:
                 warnings.warn(f"{self.__element_name} {name}[{k}] is not in the list of displayed columns")
 
     def _get_dataframe_attributes(self, date_format="MM/dd/yyyy", number_format=None) -> '_Builder': # noqa: C901
-        columns, col_types, reset_fn_hash = self.__gui._calculate_table_columns(_is_boolean_true(t.cast(bool, self.__attributes.get("reset", False))), self.__hashes.get("reset"),
+        columns, col_types, rebuild_fn_hash = self.__gui._calculate_table_columns(_is_boolean_true(t.cast(bool, self.__attributes.get("rebuild", False))), self.__hashes.get("rebuild"),
             self.__attributes.get("data"), self.__hashes.get("data", ""),
             self.__attributes.get("columns", {}), self.__hashes.get("columns"),
             _add_to_dict_and_get(self.__attributes, "date_format", date_format), _add_to_dict_and_get(self.__attributes, "number_format", number_format))
-        if reset_fn_hash:
-            self.__set_react_attribute("columns", reset_fn_hash)
+        if rebuild_fn_hash:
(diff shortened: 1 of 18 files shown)


src/taipy/gui/utils/_variable_directory.py
```diff
@@ -103,7 +103,7 @@ def get_var(self, name: str, module: str) -> t.Optional[str]:
 
 
 _MODULE_NAME_MAP: t.List[str] = []
-_RE_TPMDL_DECODE = re.compile(r"(.*?)_TPMDL_(.*)")
+_RE_TPMDL_DECODE = re.compile(r"(.*?)_TPMDL_(\d+)$")
```
dinhlongviolin1: Nice touch

src/taipy/gui/renderers/utils.py
```diff
@@ -27,6 +27,10 @@ def _get_tuple_val(attr: tuple, index: int, default_val: t.Any) -> t.Any:
     return attr[index] if len(attr) > index else default_val
 
 
+def _get_column_desc(columns: t.Dict[str, t.Any], key: str) -> t.Dict[str, t.Any]:
```
dinhlongviolin1: Much needed refactor

dinhlongviolin1 (approved):
Nicely done

src/taipy/gui/utils/_variable_directory.py
```diff
@@ -103,7 +103,8 @@ def get_var(self, name: str, module: str) -> t.Optional[str]:
 
 
 _MODULE_NAME_MAP: t.List[str] = []
-_RE_TPMDL_DECODE = re.compile(r"(.*?)_TPMDL_(.*)")
+_RE_HAS_TPMDL = re.compile(r".*_TPMDL_.*")
```
FabienLelaquais: I don't see _RE_HAS_TPMDL anywhere... and the RE seems overkill.
If you need to find the string there are more effective ways

src/taipy/gui/renderers/builder.py
```diff
@@ -464,44 +464,40 @@ def __update_col_desc_from_indexed(self, columns: t.Dict[str, t.Any], name: str)
...
-        col_types = self.__gui._accessors._get_col_types(data_hash, _TaipyData(data, data_hash))
-        columns = _get_columns_dict(
-            data,
-            _add_to_dict_and_get(self.__attributes, "columns", {}),
-            col_types,
-            _add_to_dict_and_get(self.__attributes, "date_format", date_format),
-            _add_to_dict_and_get(self.__attributes, "number_format", number_format),
-        )
+    def _get_dataframe_attributes(self, date_format="MM/dd/yyyy", number_format=None) -> '_Builder': # noqa: C901
+        columns, col_types, reset_fn_hash = self.__gui._calculate_table_columns(_is_boolean_true(t.cast(bool, self.__attributes.get("reset", False))), self.__hashes.get("reset"),
```
FabienLelaquais: Couldn't spot where _calculate_table_columns() is defined... strange

FabienLelaquais (approved):
I get the excellent idea.
Great job!

FabienLelaquais:
> via reset property (@FabienLelaquais do you have/want a better name ?)
> 
> Example: switch column order
> 
> ```
> from taipy.gui import Gui
> 
> data = {"Col 1": [f"name {i}" for i in range(1000)], "Col 2": [True if i % 2 else False for i in range(1000)]}
> 
> columns="Col 1;Col 2"
> 
> def switch_cols(state):
>     state.columns = "Col 2;Col 1" if state.columns == "Col 1;Col 2" else "Col 1;Col 2"
> 
> page = """
> <|{data}|table|reset|columns={columns}|>
> <|Switch Cols|button|on_action=switch_cols|>
> """
> 
> Gui(page=page).run()
> ```

How about "reconstruct"?

Please create another issue for me to create a full example of this, and update the ReleaseNotes

src/taipy/gui/utils/_variable_directory.py
FredLL-Avaiga: it is used in Gui but you're right it is an overkill

src/taipy/gui/renderers/builder.py
FredLL-Avaiga: in Gui

FredLL-Avaiga:
> How about "reconstruct"?
> 
Don't like it

Issue #617, opened 2023-02-22
Table improve rebuild


FredLL-Avaiga:
**Description**
Handle all truthy data for rebuild
do not update value on non rebuild
merge columns objects

**Acceptance Criteria**
- [x] Ensure new code is unit tested, and check code coverage is at least 90%
- [x] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [x] Ensure any change is well documented

PR #618, opened 2023-02-22
Improve rebuild property (#617)

FredLL-Avaiga:
Addresses #617.

handle truthy data
do not update value on no refresh
merge columns object


dinhlongviolin1 (approved):

FabienLelaquais (approved):
Well...

Issue #572, opened 2023-01-16
Taipy GUI Styling Toolbox


ghost:
### Initial finding
Taipy GUI is a great tool for making dashboards and data visualization boards but it has two style-related limitations :
- Some of the default GUI styles could be normalized to give the UI more consistency.
- Out of the box, it can be challenging for frontend beginners to customize their app's look and feel.

### Description of the solution
[taipy-gui-styling-toolbox.zip](https://drive.google.com/file/d/1ZxN0Y9HXyJI4nMSNRUbrd9wp0QXV2CMY/view?usp=share_link)

This GUI Styling Toolbox (**_GUI-ST_** ?) aims to address these issues : 
- Normalizing Taipy visual elements styles by extending the default MUI theme (in /assets/mui_theme.py) and overriding default components styles.
- Taking advantage of GUI styling capabilities (visual elements `class_name` parameter and markdown "Attribute Lists" extension) and providing the user a limited but useful set of utility styles and pure CSS components (such as header, sidebar, card) through a tiny CSS framework (`main.css`, importing CSS files from `/assets/styles`). Most of these styles are customizable through custom properties (CSS variables) to give the users even more customization flexibility and to support light/dark mode.

### Questions
- How could (and should) this feature be integrated to the product ?
- Some values, such as color values, or font-family, are duplicated from MUI theme (for MUI related components styling) to the CSS framework custom properties (that can handle every other style, like backgrounds, or additional button styles, for instance). Do you think of a way to avoid this duplication and just define these values once ?

### Acceptance Criteria
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [x] Propagate any change on the demos and run all of them to ensure there is no breaking change (done on the production planning demo)
- [ ] Ensure any change is well documented (not yet documented because it needs to be discussed before)

PR #620, opened 2023-02-22
theme & css stylekit (#572 #624)

FredLL-Avaiga:
stylekit is commanded by the stylekit bool run parameter (default: True)
configurable through optional stylekit_config run parameter
`Gui(page=page).run(stylekit=True, stylekit_variables={"color_primary": "#00FF00"})`


FabienLelaquais (changes requested):
Excellent starting point.

A few changes that I think are relevant:

- Globally rename 'toolkit' to 'stylekit', rename 'toolkit_config' to 'stylekit_variables'.
     => git mv public/toolkit public/stylekit
- stylekit=True  should be the default, That should be the future of Taipy, with as little code as possible (no stylekit=True in every Taipy GUI app).
- homogeneity: please use the same coding style for css (space before {, endling line skip, 2 or 4 space chars for indentation, etc.)
- CSS file headers to be similar as well - including license?
-  There are very generic class names everywhere... like 'container'

ghost:
@FredLL-Avaiga I think that's a great starting point indeed!
There is only one thing I would do differently, and it would be to move some theme styles from this `gui/src/themes/stylekit.ts` file directly to the relevant component's files, as styled component rules. This would be better for component specific styles to stay where they are useful rather than in a deported global file.
I can do that part if you like, in this branch or in a later one after you merge this.

@FabienLelaquais as for generic class names : "container", for example, is a name commonly used in multiple big and well known frameworks: CSS frameworks like Bootstrap or JS frameworks like Material UI (to cite just them) use this class (or component) name for that very specific case of limiting content width and centering it into the window. I get why you feel it's generic, but in the CSS/JS ecosystem, container (without affix) actually is a meaningful name.

ghost (approved):

FredLL-Avaiga:
> move some theme styles from this `gui/src/themes/stylekit.ts` file directly to the relevant component's files, as styled component rules.

Why not but this need to be activated only if window.taipyConfig.stylekit ids defined ...

ghost:
> this need to be activated only if window.taipyConfig.stylekit ids defined ...

Yes, but there are some normalization and improvement styles here that could be implemented right away, whether stylekit is ON or not, I think. Like making buttons and inputs the same height, or matching text inputs and selector styles, for obvious example.

gui/src/themes/stylekit.ts
```diff
@@ -0,0 +1,274 @@
...
+                root: {
+                    // Fill the available width
+                    display: "flex",
+
+                    "&.taipy-selector": {
+                        marginLeft: 0,
+                        marginRight: 0,
+                    },
+
+                    ".taipy-tree &": {
```
FredLL-Avaiga: @JaufreyTpy I don't understand these rules: for taipy-tree and taipy-layout, the components are not embedded in a FormControl ?
(this is valid for taipy-selector though and indeed I will apply it directly to the taipy component)

FredLL-Avaiga: oh sorry, I think I read it wrong.

FredLL-Avaiga: for taipy-tree and taipy-layout, these rules address the included FormControl (ie the TextField inside the taipy-tree Box)

FabienLelaquais:
The last commit looks neat to me: a straightforward way to get tests through.
Now... would it be possible before the merge to "git mv gui/public/stylekit/visual-elements/taipy-blocks" and "taipy-controls" to "blocks" and "controls" respectively? That prefix (even worse at this level) makes very little sense to me
Thx

src/taipy/gui/gui.py
```diff
@@ -244,8 +244,9 @@ def __init__(
         self._flask = flask
         if css_file is None:
             script_file = pathlib.Path(self.__frame.f_code.co_filename or ".").resolve()
-            css_file = f'{script_file.stem or "Taipy"}.css'
-        self._css_file = css_file
+            if script_file.with_suffix(".css").exists():
```
FabienLelaquais, line 247: Finally! Thx!

FabienLelaquais (approved):
👌

Issue #630, opened 2023-02-27
Use regular MUI <Button> component for file_download and file_selector


ghost:
Just like the `button` control, `file_download` and `file_selector` controls should use a regular MUI [Button](https://mui.com/material-ui/react-button/) component instead of the current [FAB](https://mui.com/material-ui/react-floating-action-button/) (Floating Action Button) component.

![A button](https://user-images.githubusercontent.com/115568883/221540427-b14356b7-8f4c-4d70-8240-bf2385bbdd1b.png)
_A regular button_

![Current file_download](https://user-images.githubusercontent.com/115568883/221540217-7169780d-c16b-4083-8385-a1aaf0fc1587.png)
_Current file_download_

This would be a semantic and logic issue (these buttons are not meant to be floating over content, so there is no justification to use a FAB) as well as a visual and technical consistency issue (using the same component for every button-displaying control).

FlorianJacta:
To give my inexperienced point of view, I like to have a different style of button for the download/upload button as they are other visual elements and have specific behaviors (compared to buttons).

ghost:
I understand your point of view.
Actually in web design (just like in dev, as you know), consistency is really an important point and any break in consistency should have a very good justification. 
In this case, it it not only the style which differs but also the MUI component used for rendering the button. The FAB is not destined to stay in page flow but to float over content, as a fixed element. In addition to that, using different components prevents anyone (us or the user) from styling both buttons at the same time and doubles the work if you want to harmonize things.
It is to the user to decide if they want to visually separate button concerns in their app. We should provide consistent styles by default, and let the user adapt things to his needs.
And even if you want to separate, it always good to have at least some base styles in common, and overriding just the ones you want to differenciate on one button type, going from general styles to specific tweaks and modifiers.

PR #636, opened 2023-02-28
#630 change Fab to button

FredLL-Avaiga:
improve warning messages

### DIFF
--- a/src/taipy/gui/data/content_accessor.py
+++ b/src/taipy/gui/data/content_accessor.py
@@ -18,7 +18,7 @@ from importlib import util
 from pathlib import Path
 from sys import getsizeof
 
-from ..utils import _get_non_existent_file_path
+from ..utils import _get_non_existent_file_path, _variable_decode
 
 _has_magic_module = False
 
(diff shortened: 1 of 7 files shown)


ghost (approved):
The component part seems perfectly fine to me.

src/taipy/gui/data/content_accessor.py
```diff
@@ -58,6 +58,15 @@ def __get_mime_from_file(self, path: pathlib.Path):
                 warnings.warn(f"({path}) cannot read mime type.\n{e}")
         return None
 
+    def __get_display_name(self, var_name: str) -> str:
+        if not isinstance(var_name, str):
+            return var_name
+        if var_name.startswith("_tpC_"):
+            var_name = var_name[5:]
+        if var_name.startswith("tpec_"):
```
FabienLelaquais, line 66: that would be an 'else', right?

FabienLelaquais (approved):
Looks just great!

src/taipy/gui/data/content_accessor.py
FredLL-Avaiga: no, the var can be encapsulated in a TaipyContent holder

FabienLelaquais: Oh whao. I actually couldn't figure out where the tpC part came from. Kind of searched... then abandoned.
Makes sense. Thx

Issue #632, opened 2023-02-27
Use img tag for image control instead of background-image


ghost:
On the image control: why not using an <img> tag ? Currently, the image is inserted as a span`s background-image.

Accessibility-wise, content images should not be inserted this way as screen readers can't parse it. And as a pure styling matter, it makes manipulating image dimensions a pain, as we can't just let the image resize homothetically depending on intrinsic size and container limits and flow in the content, as an img would allow it. 

In most case, using an img is the recommanded way to insert images.
And one more thing : the current background image vertical positioning of 40% feels weird. Why not centering the image in its container by default ?

FredLL-Avaiga:
trying to simplify image button with text
![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/27b54333-acb3-49f0-bce3-b18cd0a6983b)
@JaufreyTpy How would you go about making the text more visible?
it is in a span that has this style
overflow: "hidden",
    pointerEvents: "none",
    position: "absolute",
    zIndex: 0,
    top: 0,
    right: 0,
    bottom: 0,
    left: 0,
    borderRadius: "inherit",
    display: "flex",
    justifyContent: "center",
    alignItems: "center",

ghost:
Wow... like this, it feels like a very bad idea to me. Whatever the solution we use, if we lay the text over the image there is too much risk for the text to lack readbility. Not mentionning the ugliness...
I would either put the text in a tooltip or center it right under the image, as a caption.

FredLL-Avaiga:
Tooltip then

> I would either put the text in a tooltip or place it under the image.

FredLL-Avaiga:
```
from taipy.gui import Gui

my_image_file = "c:\\users\\...\\fred.png"
my_svg_file = "c:\\users\\...\\LogoTaipy.svg"
my_svg = """
<svg version="1.1"
    viewBox="0 0 300 200"
     xmlns="http://www.w3.org/2000/svg">
  <rect width="100%" height="100%" fill="red" />
  <circle cx="150" cy="100" r="80" fill="green" />
  <text x="150" y="125" font-size="60" text-anchor="middle" fill="white">SVG</text>
</svg>
"""

def on_action(state):
    print("on_action(state)")

page = """
<|toggle|theme|>
<|image|content={my_image_file}|width=unset|>
<|image|content={my_svg_file}|>
<|image|content={my_svg}|>
<|image|content={my_image_file}|label=image label|>
[...]

Issue #631, opened 2023-02-27
Add ability to declare different menu icons for dark and light mode


ghost:
Could we try to think of a way to let the user define different icons depending of the mode used ? Right now, it's not possible and black icons keep appearing black in dark mode, which makes them almost invisible.

![image](https://user-images.githubusercontent.com/115568883/221543479-be6122ce-1570-474e-a374-9be8ee35e59b.png)
_An icon menu in light mode_

![image](https://user-images.githubusercontent.com/115568883/221543595-ca5c6482-9ea9-4d2e-9af3-e17bab678295.png)
_The same icon menu in dark mode_

For now, the way to work with menu icons in multiple color modes is currently to design colored icons that can display well in both modes, or add a background right to the icon image, which is not ideal and limits styling options.

Another way of allowing icons to adapt would be to render SVG images code in the output DOM to be able to act on the SVG styles with CSS rules, from the component level or from a stylesheet.

FredLL-Avaiga:
@FlorianJacta I suppose you're the right person to ask for these icons ?

ghost:
@FredLL-Avaiga I actually have the whole icon set restyled (from taipy studio) that you can use here. I give them to you right away.

ghost:
@FredLL-Avaiga Here's the whole set in black and white PNG.
[taipycons.zip](https://github.com/Avaiga/taipy-gui/files/10851790/taipycons.zip)

FredLL-Avaiga:
Tx Jauf

FredLL-Avaiga:
@JaufreyTpy 
What do you think of
![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/7cfc9316-9b40-47d7-86d8-e5bfd5790b8d)
![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/a65131d8-d5c2-4da1-9757-8c41e56abe63)
The taipy icon is embedded as SVG

ghost:
@FredLL-Avaiga Seeing it like this, it seems way better to me. How is the SVG rendered ? In an `<img src="">` or as an inline `<svg>` tag ?

FredLL-Avaiga:
> How is the SVG rendered ? In an `<img src="">` or as an inline `<svg>` tag ?

The content of the svg file is dumped inside the avatar div => svg tag

PR #637, opened 2023-02-28
#631 base image

FredLL-Avaiga:
#632 themed and svg icons
#638 Inline SVG

New optional properties in Icon
- light_path: str
- dark_path: str

### DIFF
--- a/gui/src/components/Taipy/Image.tsx
+++ b/gui/src/components/Taipy/Image.tsx
@@ -11,7 +11,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-import React, { CSSProperties, useCallback, useContext, useEffect, useMemo, useRef } from "react";
+import React, { useCallback, useContext, useEffect, useMemo, useRef } from "react";
 import axios from "axios";
 import Button from "@mui/material/Button";
 import Tooltip from "@mui/material/Tooltip";
@@ -30,21 +30,6 @@ interface ImageProps extends TaipyActiveProps {
     content?: string;
     defaultContent: string;
 }
-
-const labelSpanStyle = {
-    overflow: "hidden",
-    pointerEvents: "none",
-    position: "absolute",
-    zIndex: 0,
-    top: 0,
-    right: 0,
-    bottom: 0,
-    left: 0,
-    borderRadius: "inherit",
-    display: "flex",
-    justifyContent: "center",
-    alignItems: "center",
-} as CSSProperties;
 
 const Image = (props: ImageProps) => {
     const { id, onAction, width = 300, height } = props;
--- a/gui/src/components/Taipy/Image.tsx
+++ b/gui/src/components/Taipy/Image.tsx
@@ -11,7 +11,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-import React, { useCallback, useContext,  useMemo } from "react";
+import React, { useCallback, useContext, useMemo } from "react";
 import { styled, SxProps, Theme } from "@mui/material/styles";
 import ButtonBase from "@mui/material/ButtonBase";
 import Typography from "@mui/material/Typography";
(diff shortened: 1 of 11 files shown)


dinhlongviolin1 (approved):

src/taipy/gui/icon.py
```diff
@@ -45,14 +46,29 @@ def __init__(self, path: str, text: t.Optional[str] = None) -> None:
                 the application directory.
             text (Optional[str]): The text associated to the image. If _text_ is None,
                 there is no text associated to this image.
+            light_path (Optional[str]): The path to the light theme image (fallback to _path_ if not defined).
+            dark_path (Optional[str]): The path to the dark theme image (fallback to _path_ if not defined).
+            svg (Optional[bool]): Indicates that the _path_ should be treated as an SVG resource (defaults to True if _path_ ends with .svg).
```
FabienLelaquais: Where is this doc ("defaults to True if _path_ ends with .svg") implemented?
What if path is svg, but light_path is not?

src/taipy/gui/icon.py
```diff
@@ -45,14 +46,29 @@ def __init__(self, path: str, text: t.Optional[str] = None) -> None:
                 the application directory.
             text (Optional[str]): The text associated to the image. If _text_ is None,
                 there is no text associated to this image.
+            light_path (Optional[str]): The path to the light theme image (fallback to _path_ if not defined).
```
FabienLelaquais: Please use '*' for parameter name references instead of '_'.

gui/src/components/Taipy/Image.tsx
```diff
@@ -129,25 +129,35 @@ const Image = (props: ImageProps) => {
...
+                                className={getSuffixedClassNames(className, "-label")}
+                            >
+                                {label}
+                                <ImageMarked className="MuiImageMarked-root" />
+                            </Typography>
+                        </ImageSpan>
+                    )}
+                </ImageButton>
+            ) : (
+                <img id={id} src={content} style={style} className={className} />
```
FabienLelaquais: So the potential svg is embedded in an img... hmm ok...

FabienLelaquais (approved):
Nice job!

src/taipy/gui/icon.py
FredLL-Avaiga: In [icon.tsx:57](https://github.com/Avaiga/taipy-gui/blob/feature/%23631-%23632-base-image-and-themed-icons/gui/src/utils/icon.tsx#L57)
the svg property is meant to allow non svg ending files to be treated as SVG ... Is it really necessary?

src/taipy/gui/icon.py
FredLL-Avaiga: will do

gui/src/components/Taipy/Image.tsx
FredLL-Avaiga: Didn't think of this one ...

ghost: Yeah I was not sure I read it well, but it tickled me as well...

FredLL-Avaiga: We could also handle content as svg text ie 
```
content = "<SVG> ...</SVG>"
<|image|content={content}|>
```
What do you think ?

ghost: You meam, right in a GUI app, on the user side ? It's a bit dirty, don't you think ? SVG code can be messy, and putting that in your app's code may make it look like crap.

FredLL-Avaiga: just another option ... it can be loaded from a file or returned by a function ...

ghost: OK, why not, then.

FredLL-Avaiga: one could use this to insert svg generated from template

src/taipy/gui/icon.py
FabienLelaquais: Oh right.
And as you say... we may not even need this.

gui/src/components/Taipy/Image.tsx
FabienLelaquais: ... and animate it!

Issue #642, opened 2023-03-01
BUG-Image handle svg with size


FredLL-Avaiga:
**Description**
When displaying a svg resource where the size is defined (width height), the image should be constrained to the defined size

**How to reproduce**
use a SVG with width height
ex: https://upload.wikimedia.org/wikipedia/commons/7/79/Flag_of_Nigeria.svg

**Expected behavior**
svg should be displayed with the user defined size

PR #644, opened 2023-03-01
Handle svg with size (#642)

FredLL-Avaiga:
Addresses #642.


FabienLelaquais (approved):
Let the Flex be with you.

Issue #443, opened 2022-10-03
Dynamic property for traces - Charts


FlorianJacta:
**Description**
Sometimes, the number of traces in a Dataframe is not known. Thus, it is compulsory to create a Partial to handle it. It is making it a bit complex to use charts this way. The goal would be to simplify this use case of charts.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FlorianJacta:
We could also extend it to Tables. It seems that there is a need for it.

[Question on the forum on this matter for tables](https://www.taipy.io/forums/topic/reloading-context/)

FredLL-Avaiga:
rebuild for both

PR #648, opened 2023-03-06
rebuild chart (#443)

FredLL-Avaiga:
And improve rebuild table

### DIFF
--- a/src/taipy/gui/utils/clientvarname.py
+++ b/src/taipy/gui/utils/clientvarname.py
@@ -9,8 +9,6 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-import typing as t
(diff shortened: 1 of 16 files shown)


dinhlongviolin1 (approved):
Nicely done!

src/taipy/gui/utils/clientvarname.py
```diff
@@ -9,11 +9,20 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import typing as t
```
FabienLelaquais: Do you need that?

FabienLelaquais (approved):
Not pretending I get it all...
Looks really neat!

src/taipy/gui/utils/clientvarname.py
FredLL-Avaiga: Good catch

Issue #633, opened 2023-02-27
Table control: render `class_name` parameter on highest level.


ghost:
On the `<|table|>` element there is something not so great regarding the way class_names are rendered.
Currently, `class_name` applies to the html `<table>` tag, two levels deep inside the component.
Instead, it should render on the highest element too, because if not, using class names on it can produce unintended effects.

It is the case with margins.
The intended result of applying a margin-top is to space our `<|table|>` from the element above. And in CSS, if that previous element has a margin-bottom of its own, the margins of two adjacent siblings should collapse. Instead, as we apply our margin to a child of our table, the margins add to one-another.
Same goes if we want to just nullify our `<|table|>` default margin-top with a `.m0` utility class, for example. Currently, we can't, because the default margin is on the highest level and the class_name is referencing a child level.

The general rule for dynamic classes should be to always apply to the highest level as CSS can target a child element going down from its parent's name but (almost) not the other way around.

ghost:
@FredLL-Avaiga 
I see you made a change, here, but the `taipy-table` class is still applied to the `<table>` element instead of the `<div class="MuiBox-root">` right under the `md-para` element (as shown on the image bellow).
![image](https://user-images.githubusercontent.com/115568883/224684717-5de47563-3e3f-4d8f-b67a-863e7eef99aa.png)
It's still not good, on a CSS styling point of view. Is it what you intended to do with your last fix, or is it a problem on my side of the lib ?

FredLL-Avaiga:
Is it possible that you're not checking the right version ?
Here is the code for PaginatedTable.tsx
```
return (
        <Box id={id} sx={boxSx} className={`${className} ${getSuffixedClassNames(className, "-paginated")}`}>
...
```

ghost:
Unfortunately not. My branch is up to date with develop, and I have exactly this in the component... but still, it renders as I told you above. 
I rebuilt the src/taipy/gui app to be sure, but it has no effect either. 
Here is my table snippet in Gui : 
`<|{dataset_week}|table|width=100%|on_edit=edit_function|>`

FredLL-Avaiga:
I suppose You have an issue with your build.
When I show a table, here is what I get:
```
<div class="md-para">
<div class="taipy-table taipy-table-paginated MuiBox-root css-306mec">
<div class="MuiPaper-root ...
```

ghost:
[...]

PR #655, opened 2023-03-10
#633 classname on top level element

FredLL-Avaiga:


FabienLelaquais (approved):
I like that.
There's a slight risk of CSS selectors not working properly now right? Release Notes is unfortunately unavoidable.

FredLL-Avaiga:
> Release Notes is unfortunately unavoidable.
Should we add a specific label?

FabienLelaquais:
> 

I suspect the best would be to have a working document, not stored in GitHub (too bad), where we keep track of that...

Issue #612, opened 2023-02-20
BUG- Map not loading in multi-page application


FlorianJacta:
**Description**
A strange behavior has appeared in Taipy 2.1 (not present in Taipy 2.0). Loading a map on a page doesn't work sometimes, markers are not displayed. It seems to happen when the application is multi-page and when the user goes from one page to the other.

**How to reproduce**
You should land on the chart page. Going from one page to another, no markers are present on the map, even if marker texts can be found on the map. Refreshing the page will show the markers again.

```python
from taipy.gui import Gui 

data = {"number":[50,78,95,78,12,46],
        "size":[50,78,95,63,45,23],
        "lat":[39.83049768,49.83049768,59.83049768,69.83049768,35.83049768,33.83049768],
        "lon":[39.83049768,49.83049768,59.83049768,69.83049768,35.83049768,33.83049768],}

marker_map = {"color":"number", "size": "size"}
layout_map = {"mapbox": { "style": "open-street-map"}}

map = "<|{data}|chart|type=scattermapbox|lat=lat|lon=lon|marker={marker_map}|layout={layout_map}|mode=markers|height=800px|>"

pages = {"/":"<|navbar|>",
        "chart":"# Chart",
         "map":map}

Gui(pages=pages).run()
```

**Expected behavior**
The map should always show the markers directly going from one page to the other.

**Screenshots**
![map](https://user-images.githubusercontent.com/98709993/220065662-77af8c52-c59b-4bb1-95fd-bc2da587a809.gif)

**Runtime environment**
Please specify relevant indications.
 - Taipy GUI 2.1.0 (working in Taipy GUI 2.0)

FredLL-Avaiga:
To be oinvestigated but looks like a plotly issue to me (if the text markers are visible) ...

FredLL-Avaiga:
Looks like an issue between plotly and mapboxGL
https://stackoverflow.com/questions/68562009/what-does-error-there-is-already-a-source-with-this-id-anonymous-function-m

FredLL-Avaiga:
the browser console shows:
![image.png](https://images.zenhubusercontent.com/613efb225a9c163a3da6610f/92e72d6f-ae50-4306-800c-f1654ed98ef1)

FredLL-Avaiga:
I guess we'll have to wait for plotly to fix this ...

FlorianJacta:
[...]

PR #656, opened 2023-03-10
chart refresh (#612)

FredLL-Avaiga:
DataKey management : to allow for a better refresh

### DIFF
--- a/gui/src/components/Taipy/Chart.tsx
+++ b/gui/src/components/Taipy/Chart.tsx
@@ -420,9 +420,7 @@ const Chart = (props: ChartProp) => {
                         .map(([k, v]) => `${k}=${v}`)
                         .join("-");
                     const dtKey = backCols.join("-") + (config.decimators ? `--${config.decimators.join("")}` : "") + "--" + eventDataKey;
-                    if (dataKey !== dtKey) {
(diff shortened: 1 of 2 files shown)


gui/src/components/Taipy/Chart.tsx
```diff
@@ -246,18 +246,23 @@ const Chart = (props: ChartProp) => {
...
-                createRequestChartUpdateAction(
-                    updateVarName,
-                    id,
-                    backCols,
-                    dataKey.current,
-                    getDecimatorsPayload(config.decimators, plotRef.current, config.modes, config.columns, config.traces)
-                )
-            );
+            const dtKey = backCols.join("-") + (config.decimators ? `--${config.decimators.join("")}` : "");
+            if (dataKey !== dtKey) {
```
FabienLelaquais: I thought that the hook would take care of that...

FabienLelaquais (approved):
If it works... it works.
That would be a 2.1 hotfix I suspect.

gui/src/components/Taipy/Chart.tsx
FredLL-Avaiga: Good catch, thank you

PR #658, opened 2023-03-13
feat(stylekit): improve table style

ghost:
Improve `<|table|>` control styles with stylekit defaults.

### DIFF
--- a/gui/src/components/Taipy/tableUtils.tsx
+++ b/gui/src/components/Taipy/tableUtils.tsx
@@ -128,7 +128,7 @@ export const baseBoxSx = { width: "100%" };
 export const paperSx = { width: "100%", mb: 2 };
 export const tableSx = { minWidth: 250 };
 export const headBoxSx = { display: "flex", alignItems: "flex-start" };
-export const iconInRowSx = { fontSize: "body2.fontSize" };
+export const iconInRowSx = { gridColumnStart: 2, fontSize: "body2.fontSize" };
 const cellBoxSx = { display: "grid", gridTemplateColumns: "1fr auto", alignItems: "center" } as CSSProperties;
 const tableFontSx = { fontSize: "body2.fontSize" };
 
--- a/gui/src/components/Taipy/tableUtils.tsx
+++ b/gui/src/components/Taipy/tableUtils.tsx
@@ -197,7 +197,7 @@ const getCellProps = (col: ColumnDesc, base: Partial<TableCellProps> = {}): Part
     switch (col.type) {
         case "int":
         case "float":
-            base.align = "right";
+            base.align = "left";
             break;
         case "bool":
             base.align = "center";
--- a/gui/src/components/Taipy/tableUtils.tsx
+++ b/gui/src/components/Taipy/tableUtils.tsx
@@ -195,10 +195,6 @@ const renderCellValue = (val: RowValue | boolean, col: ColumnDesc, formatConf: F
 
 const getCellProps = (col: ColumnDesc, base: Partial<TableCellProps> = {}): Partial<TableCellProps> => {
     switch (col.type) {
-        case "int":
-        case "float":
-            base.align = "left";
-            break;
         case "bool":
             base.align = "center";
             break;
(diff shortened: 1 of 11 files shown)


gui/src/themes/stylekit.ts
```diff
@@ -42,6 +42,15 @@ export const stylekitTheme = {
     },
     // Components normalization
     components: {
+        // Form control
```
FredLL-Avaiga: bad comment ?

FredLL-Avaiga (commented):
Did you test the impact of your changes on autoloading table ?
Maybe use taipy-table-paginated ?

gui/src/themes/stylekit.ts
ghost: Fixed comment

ghost:
@FredLL-Avaiga Right. I fixed a weird right alignment on numbers in autoloading tables. Seems fine now.

gui/src/components/Taipy/tableUtils.tsx
```diff
@@ -197,7 +197,7 @@ const getCellProps = (col: ColumnDesc, base: Partial<TableCellProps> = {}): Part
     switch (col.type) {
         case "int":
         case "float":
-            base.align = "right";
```
FredLL-Avaiga, line 200: @JaufreyTpy Here you change numeric value alignment for all columns with or without styleKit.
If it's really what's wanted, I guess we should remove the int and floats case (L 198->201)

ghost: @FredLL-Avaiga  Yeah, that's wanted, but I wasn't sure of what I was doing there, so I left it for the moment.
I'll remove it all right, because I can't see a reason why we would want to mess with alignment on numbers like before. The more normalized it is, the easiest it is to interact with it.

Btw, is the
```
case "bool":
            base.align = "center";
            break;
```
for centering a toggle button (switch) or something like that ? I don't know how to test that case.

FredLL-Avaiga: create a column of boolean values

FredLL-Avaiga (approved):

gui/src/components/Taipy/tableUtils.tsx
```diff
@@ -128,7 +128,7 @@ export const baseBoxSx = { width: "100%" };
 export const paperSx = { width: "100%", mb: 2 };
 export const tableSx = { minWidth: 250 };
 export const headBoxSx = { display: "flex", alignItems: "flex-start" };
-export const iconInRowSx = { fontSize: "body2.fontSize" };
+export const iconInRowSx = { gridColumnStart: 2, fontSize: "body2.fontSize" };
```
FredLL-Avaiga: Did you check the impacts of this change ?
- Add a row icon
- Aggregate icon
- Table filter icon
- Table Filter list row Validate and Delete icons
- Table value edit icons

ghost: - Add a row icon: fixed
- Aggregate icon: fixed 
- Table filter icon: check
- Table Filter list row Validate and Delete icons: fixed
- Table value edit icons: check

I took this occasion to improve the filter popover style too.

gui/src/themes/stylekit.ts
```diff
@@ -173,6 +175,16 @@ export const stylekitModeThemes = {
             },
         },
         components: {
+            // Give popover paper a slighly lighter color to reflect superior elevation
+            MuiPopover: {
+                styleOverrides: {
+                    paper: {
+                        backgroundColor: !!window?.taipyConfig?.stylekit?.colorPaperLight
```
FredLL-Avaiga: !!window?. => window.

FredLL-Avaiga (approved):

gui/src/themes/stylekit.ts
ghost: @FredLL-Avaiga Woops, indeed, that was overkill. Fixed.

FabienLelaquais (approved):
👍

PR #664, opened 2023-03-14
fix(stylekit): apply CSS vars to root selector

ghost:
Best practice: Apply CSS vars to root selector to decrease specificity.

### DIFF
--- a/gui/public/index.html
+++ b/gui/public/index.html
@@ -12,11 +12,13 @@
     {%- if version -%}
     window.taipyVersion="{{version}}";
     {% endif %}</script>
-    <style>
-      :root{
-        {{css_vars}}
-      }
-    </style>
+    {%- if css_vars -%}
+      <style>
+        :root{
+          {{css_vars}}
+        }
+      </style>
+    {% endif %}
     <!-- Taipy Gui js <%= htmlWebpackPlugin.options.frontend_version %> <%= htmlWebpackPlugin.options.frontend_build_date %> <%= htmlWebpackPlugin.options.frontend_build_mode %> -->
   </head>
   <body >
--- a/gui/public/index.html
+++ b/gui/public/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" style="{{css_vars}}">
+<html lang="en">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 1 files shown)


gui/public/index.html
```diff
@@ -12,6 +12,11 @@
     {%- if version -%}
     window.taipyVersion="{{version}}";
     {% endif %}</script>
+    <style>
```
FredLL-Avaiga: interesting, I have often see the pattern &lt;html style="vars "> (by Microsoft and others ...)

FredLL-Avaiga (dismissed):

FredLL-Avaiga (commented):
We should not allow user override of these css vars as this would bring inconsistency between CSS and MUI Theme.
We want users to change those css vars via the stylekit_variables in Gui Config.

gui/public/index.html
ghost: It's not uncommon, indeed, but it makes overriding CSS custom properties a little less easy and intuitive.
The good practice in CSS for them is to assign base variables to the :root selector. Problem is this selector has a lower priority than tag selectors like `<html>` so in order to override them, you have to target the html selector, which is counter-intuitive, especially for beginers. 
I'd say, let's set the lowest CSS priority on our stylekit variables so whatever the selector used by the user to reassign them, he/she will see his/her own taken into account instead.

ghost:
@FredLL-Avaiga I understand your point, but using CSS variables allows motivated users to do redefine them, whether you want it or not, by scoping a descendant selector like `body`. Fighting that is fighting windmills. Or you would have to assign these variables as inline styles to each and every html element in the DOM to ensure that users don't play with it. And even like this, clever ones would overcome.
This :root selector thing works just as well, but avoids creating frustration and shows we follow best practices.

FredLL-Avaiga:
You're right I don't think we should fight users :-) or windmills !
But we should promote the use of stylekit_variables as a way of modifying those css vars.
From your change, I understood that it allows easier overriding of these and thus easier inconsistency between pure CSS and MUI Theme ?

ghost:
@FredLL-Avaiga In that I agree with you, we don't have to promote this. My initial argument was kinda wrong and I stand corrected. 
But I still think it's good to apply this change as a best practice to let the door open.

gui/public/index.html
FredLL-Avaiga: Can you embed this in 
```
{%- if css_vars -%}
<style>
...
{% endif %}
```
?

FredLL-Avaiga (approved):

gui/public/index.html
ghost: Of course. Done !

FabienLelaquais (approved):
After those discussions... Ok!

PR #665, opened 2023-03-15
fix(stylekit): get rid of unneeded image styles

ghost:
Ditch styles which had not point anymore since the Image component had been changed, because they mostly acted on background based images.
Fix default inline images alignment.

### DIFF
--- a/gui/public/index.html
+++ b/gui/public/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en" style="{{css_vars}}">
+<html lang="en">
   <head>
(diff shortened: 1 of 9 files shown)


gui/public/index.html
FredLL-Avaiga, line 1: No need for this

ghost: Woops, I don't know what that was doing here. I probably didn't create this branch on develop.
That's fixed, I reverted that commit.

FabienLelaquais (approved):
👌

FredLL-Avaiga (approved):

Issue #660, opened 2023-03-13
Stylekit: Give `borderRadius` and `inputButtonHeight` variables the `string` type in the StyleKitVars interface


ghost:
I came across the `gui/src/utils/index.ts` files `StyleKitVars` interface here:

```
interface StyleKitVars {
    // Primary and secondary colors
    colorPrimary: string;
    colorSecondary: string;
    // Contextual color
    colorError: string;
    colorWarning: string;
    colorSuccess: string;
    // Background and elevation color for LIGHT MODE
    colorBackgroundLight: string;
    colorPaperLight: string;
    // Background and elevation color for DARK MODE
    colorBackgroundDark: string;
    colorPaperDark: string;
    // DEFINING FONTS
    // Set main font family
    fontFamily: string;
    // DEFINING SHAPES
    // Base border radius
    borderRadius: number;
    // DEFINING MUI COMPONENTS STYLES
    // Matching input and button height
    inputButtonHeight: number;
}
```

And I wonder if, for the reasons we discussed before, we shouldn't give `borderRadius` and `inputButtonHeight` the `string` type here.

What do you think ?

_Originally posted by @JaufreyTpy in https://github.com/Avaiga/taipy-gui/issues/650#issuecomment-1466369547_

FredLL-Avaiga:
setting theme borderRadius to a string is not supported by Mui
`MUI: The `theme.shape.borderRadius` value (8px) is invalid.`

ghost:
Damned...
Then, if the MUI value has to be a number, we'll have to insert the `px` unit after the value in the CSS custom prop definition, because a unitless number won't work with CSS.
And I'll have to change this statement in the documentation too.

FredLL-Avaiga:
I suppose input_button_height can stay as a string

ghost:
Yes, this one can stay as it is.

PR #671, opened 2023-03-17
styleKit  vars: border_radius (#660)

FredLL-Avaiga:
border_radius: int in theme and str with px in CSS
input button height: str
fix " issue in HTML :root
Type the stylekit variables dict

### DIFF
--- a/gui/src/utils/index.ts
+++ b/gui/src/utils/index.ts
@@ -47,11 +47,11 @@ interface StyleKitVars {
     // Set main font family
     fontFamily: string;
     // DEFINING SHAPES
-    // Base border radius
+    // Base border radius in px
     borderRadius: number;
     // DEFINING MUI COMPONENTS STYLES
     // Matching input and button height
-    inputButtonHeight: number;
+    inputButtonHeight: string;
 }
 
 // return client server timeZone offset in minutes
--- a/gui/src/utils/index.ts
+++ b/gui/src/utils/index.ts
@@ -50,7 +50,7 @@ interface StyleKitVars {
     // Base border radius in px
     borderRadius: number;
     // DEFINING MUI COMPONENTS STYLES
-    // Matching input and button height in css size unit
+    // Height in css size unit for inputs and buttons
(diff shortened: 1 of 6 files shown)


gui/src/utils/index.ts
```diff
@@ -47,11 +47,11 @@ interface StyleKitVars {
     // Set main font family
     fontFamily: string;
     // DEFINING SHAPES
-    // Base border radius
+    // Base border radius in px
```
FabienLelaquais, line 50: I love that you mention the unit we're dealing with.
I wish that would appear everywhere (like in the inputButtonHeight property right after)

FabienLelaquais (approved):
Looks good!

gui/src/utils/index.ts
FredLL-Avaiga: inputButtonHeight is a string and the unit can be anything

FabienLelaquais: That's my point. CSS unit I guess.

FredLL-Avaiga: added in css size unit

FredLL-Avaiga: Not sure about the wording, if you have a better proposal ?

FabienLelaquais: In my opinion the wording should be the one we've used in the elements' doc.
"The &lt;whatever&gt;, in CSS units, of this &lt;bla&gt;;."

src/taipy/gui/_default_config.py
```diff
@@ -62,11 +62,11 @@
         "font_family": "Lato, Arial, sans-serif",
 
         # DEFINING SHAPES
-        # Base border radius
-        "border_radius": "8px",
+        # Base border radius in px
+        "border_radius": 8,
```
dinhlongviolin1, line 66: Thanks for this addition. TS has been screaming at me on the studio side on this value

dinhlongviolin1 (approved):

Issue #628, opened 2023-02-27
BUG-Weakness when parsing unknown elements in Markdown


FabienLelaquais:
**Description**

Unknown elements are parsed as 'text' with no warning, resulting on not knowing what's going wrong.

**How to reproduce**

Consider the following code:

```
from taipy.gui import Gui

page = """  
# Unknown element

Control: <|Some text|foobar|>
"""

Gui(page).run()
```

This results on a page that shows the title, and "Some text" just like if the construct was `<|Some text|>`.

**Expected behavior**

At least a warning should be issues. Better would be a visual feedback on the page itself, just like when a extension elements cannot be rendered.

PR #689, opened 2023-03-29
Enhance Markdown Parsing For Invalid Control (#628)

dinhlongviolin1:
Solve Issue #628
```python
from src.taipy.gui import Gui

page = """  
# Unknown element

Control: <|Some text|foobar|>
"""

Gui(page).run()
```

### DIFF
--- a/src/taipy/gui/renderers/_markdown/preproc.py
+++ b/src/taipy/gui/renderers/_markdown/preproc.py
@@ -19,6 +19,9 @@ from markdown.preprocessors import Preprocessor as MdPreprocessor
 from ..builder import _Builder
 from .factory import _MarkdownFactory
 
+if t.TYPE_CHECKING:
+    from ...gui import Gui
+
 
 class _Preprocessor(MdPreprocessor):
     # ----------------------------------------------------------------------
@@ -56,6 +59,8 @@ class _Preprocessor(MdPreprocessor):
     #  Note 1: 'not <prop_name>=<prop_value>' is an invalid syntax
     #  Note 2: Space characters after the equal sign are significative
     __PROPERTY_RE = re.compile(r"((?:don'?t|not)\s+)?([a-zA-Z][\.a-zA-Z_$0-9]*(?:\[(?:.*?)\])?)\s*(?:=(.*))?$")
+
+    _gui: "Gui"
 
     @staticmethod
     def extend(md, gui, priority):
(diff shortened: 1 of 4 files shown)


FabienLelaquais (approved):
Hum that was more impactful than I expected!
Well done!

FredLL-Avaiga (approved):
Waow !!!
Any impact on the use of library elements without specifying the library ?

dinhlongviolin1:
> Waow !!! Any impact on the use of library elements without specifying the library ?

I am not sure what you mean but I dont think there would be any impact. What I have change mainly is on the preprocessor in case there is no control type found, then I try to resolve whether it is a text or not based on the first property.

FredLL-Avaiga:
> > Waow !!! Any impact on the use of library elements without specifying the library ?
> 
> I am not sure what you mean but I dont think there would be any impact. What I have change mainly is on the preprocessor in case there is no control type found, then I try to resolve whether it is a text or not based on the first property.

Maybe I'm wrong but would the warning show when we're using a (library.)?element  ?

dinhlongviolin1:
> > > Waow !!! Any impact on the use of library elements without specifying the library ?
> > 
> > 
> > I am not sure what you mean but I dont think there would be any impact. What I have change mainly is on the preprocessor in case there is no control type found, then I try to resolve whether it is a text or not based on the first property.
> 
> Maybe I'm wrong but would the warning show when we're using a (library.)?element ?

Can you help me testing that

FredLL-Avaiga:
> > > > Waow !!! Any impact on the use of library elements without specifying the library ?
> > > 
> > > 
> > > I am not sure what you mean but I dont think there would be any impact. What I have change mainly is on the preprocessor in case there is no control type found, then I try to resolve whether it is a text or not based on the first property.
> > 
> > 
> > Maybe I'm wrong but would the warning show when we're using a (library.)?element ?
> 
> Can you help me testing that

Here you are:
```
from taipy.gui import Gui
from taipy.gui.extension import ElementLibrary, Element, ElementProperty, PropertyType

import typing as t

md = """
# Custom test

<|Here is an html title|taipy_extension_example.title|>

<|Here is an html title with no lib|title|>

"""

def render_xhtml(properties: t.Dict[str, t.Any]):
    return f'<div><h1>{properties.get("value", "Default text")}</h1></div>'


class MyLibrary(ElementLibrary):

    elts = {
        "title": Element("value", {
            "value": ElementProperty(PropertyType.string),
        }, "h1", render_xhtml=render_xhtml),
    }

    def get_name(self) -> str:
        return "taipy_extension_example"

    def get_elements(self) -> t.Dict[str, Element]:
        return MyLibrary.elts


Gui.add_library(MyLibrary())
Gui(md).run(debug=True)

```

Issue #692, opened 2023-04-03
Context: improve management


FredLL-Avaiga:
**Description**
Allow several back-end execution contexts to co-exist in the same page

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

Issue #615, opened 2023-02-20
Add on_change property on expandable


FlorianJacta:
**Description**
The goal is to detect when an expandable is being open or closed. A Taipy user asked it; it might be interesting to include this property to this block.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented
- [ ] New entry in the release notes

PR #693, opened 2023-04-03
manage several backend execution contexts in the same page (#692)

FredLL-Avaiga:
#615 expandable on_change only if expanded property is bound
+ update dependencies


dinhlongviolin1 (approved):
Straight forward implementation. I love it very much

FabienLelaquais (approved):
No useful comment.
Great job!

Issue #695, opened 2023-04-04
BUG-App does not show on reload


dinhlongviolin1:
**Description**
Taipy GUI App sometimes does not show up on reload/page refresh. This might be because of websocket connection not being established correclty

PR #696, opened 2023-04-04
properly handle ws connection (#695)

dinhlongviolin1:
Solve bug #695

### DIFF
--- a/gui/src/context/taipyReducers.ts
+++ b/gui/src/context/taipyReducers.ts
@@ -273,6 +273,14 @@ export const initializeWebSocket = (socket: Socket | undefined, dispatch: Dispat
                 socket.connect();
             }, 1000);
         });
+        // try to reconnect on server disconnection
+        socket.on("disconnect", (reason) => {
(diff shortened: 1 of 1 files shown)


FabienLelaquais (approved):
Looks perfect to my brain

gui/src/context/taipyReducers.ts
```diff
@@ -264,8 +267,16 @@ export const initializeWebSocket = (socket: Socket | undefined, dispatch: Dispat
             sendWsMessage(socket, "ID", "TaipyClientId", id, id);
             dispatch({ type: Types.SocketConnected });
         });
+        // try to reconnect on connect_error
+        socket.on("connect_error", () => {
+            setTimeout(() => {
+                socket.connect();
```
FredLL-Avaiga, line 273: I suppose this will generate another 'connect' event ?
Should we then test the isSocketConnected in the connect event listener ?

FredLL-Avaiga (approved):
Right way to do it

PR #700, opened 2023-04-07
Documentation for the Stylekit (part 1)

FabienLelaquais:
Addresses [taipy-doc/#398](https://github.com/Avaiga/taipy-doc/issues/398)

- Add a Styling section to all elements.
- Use a single 'stylekit'  parameter to Gui.run() to enable, disable and configure the Stylekit.
- Replace underscores in Stylekit classes by hyphens.
- Rename StylekitVariables to Stylekit.

### DIFF
--- a/gui/doc/file_download.md
+++ b/gui/doc/file_download.md
@@ -10,6 +10,11 @@ Allows downloading of a file content.
     
 The download can be triggered when clicking on a button, or can be performed automatically.
 
+## Styling
+
+All the file download controls are generated with the "taipy-file_download" CSS class. You can use this class
+name to select the file download controls on your page and apply style.
+
 ## Usage
 
 ### Default behavior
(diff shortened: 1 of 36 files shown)


gui/doc/file_download.md
```diff
@@ -10,6 +10,11 @@ Allows downloading of a file content.
     
 The download can be triggered when clicking on a button, or can be performed automatically.
 
+## Styling
+
+All the file download controls are generated with the "taipy-file_download" CSS class. You can use this class
```
FredLL-Avaiga, line 15: Should wear change the _ to -?

FredLL-Avaiga (approved):

gui/doc/file_download.md
FabienLelaquais: Well I don't think so for two reasons:
- The pattern is "taipy-" (which is compatible with CSS best practices) and the name of the control (which makes it easy to remember)
- We've done this since 1.0.

dinhlongviolin1 (approved):
Looks good to me

PR #701, opened 2023-04-09
pyi file to support suggestion for gui run parameters

dinhlongviolin1:
Solve issue #715
Run this file to build the pyi file
```python
import os

os.system('stubgen ./src/taipy/gui/gui.py --no-import --parse-only --include-private --export-less -o ./')

from src.taipy.gui.config import Config

gui_config = "".join(
    f", {k}: {v.__name__} = ..." if "<class" in str(v) else f", {k}: {str(v).replace('typing', 't')} = ..."
    for k, v in Config.__annotations__.items()
)

replaced_content = ""
with open("./src/taipy/gui/gui.pyi", "r") as file:
    for line in file:
        if "def run(" in line:
            line = line.replace(", run_server: bool = ..., run_in_thread: bool = ..., async_mode: str = ..., **kwargs", gui_config)
        replaced_content = replaced_content + line

with open("./src/taipy/gui/gui.pyi", "w") as write_file:
    write_file.write(replaced_content)

```


FredLL-Avaiga (approved):
I really like the pyi file with explicit config
But I much less like the unnecessary changes in gui.py ;-)
I guess the only necessary change is the addition on StylekitVariables ?

dinhlongviolin1:
> I really like the pyi file with explicit config But I much less like the unnecessary changes in gui.py ;-) I guess the only necessary change is the addition on StylekitVariables ?

You are right. the only addition to the gui.py file is StylekitVariables. The others were made by the precommit when I pushed the code.

FredLL-Avaiga:
Did that mean that you're pre-commit hooks are not set like the ones in
GitHub?

FabienLelaquais (approved):
👌

FabienLelaquais:
> 
So I approved the PR with no comment, since I think that's good. The Stylekit references will change slightly, which is not a problem.

Now this pre-commit/formatting topic appears regularly and I wish we can converge to something that just works, always.
This should not be a big deal...

PR #702, opened 2023-04-10
enforce lint check correctly

dinhlongviolin1:
Solve issue #714

### DIFF
--- a/src/taipy/__init__.py
+++ b/src/taipy/__init__.py
@@ -13,11 +13,17 @@ from importlib.util import find_spec
 
 if find_spec("taipy"):
     if find_spec("taipy.config"):
-        from taipy.config import Config
-        from taipy.config import Scope
-        from taipy.config import Frequency
+        from taipy.config import Config, Frequency, Scope
 
     if find_spec("taipy.core"):
+        from taipy.core._core import Core
+        from taipy.core.common.alias import CycleId, DataNodeId, JobId, PipelineId, ScenarioId, TaskId
+        from taipy.core.cycle.cycle import Cycle
+        from taipy.core.data.data_node import DataNode
+        from taipy.core.job.job import Job
+        from taipy.core.job.status import Status
+        from taipy.core.pipeline.pipeline import Pipeline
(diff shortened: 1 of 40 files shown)


src/taipy/__init__.py
```diff
@@ -13,11 +13,17 @@
 
 if find_spec("taipy"):
     if find_spec("taipy.config"):
-        from taipy.config import Config
-        from taipy.config import Scope
-        from taipy.config import Frequency
+        from taipy.config import Config, Frequency, Scope
```
FabienLelaquais, line 16: So this file is shared (copied) across all repositories.
Any change here has to be done... at the source location.
Please check with JR where we are at this point.

src/taipy/gui/config.py
```diff
@@ -77,9 +77,9 @@
         "color_paper_dark": str,
         "font_family": str,
         "border_radius": int,
-        "input_button_height": str
+        "input_button_height": str,
```
FabienLelaquais: How come this additional (and useless) comma becomes mandatory?

FabienLelaquais (approved):
I like that a lot!
Not so much the trailing useless commas in iterators...
Well done.

src/taipy/__init__.py
FredLL-Avaiga: I guess the linter doesn't know about JR :-)
it might be a pb if we don't have the same rules and strategy for different repositories ...

src/taipy/gui/config.py
FredLL-Avaiga: Can this be configured?

FredLL-Avaiga (approved):
Not a big fan of this, but if it is for the common good ...

src/taipy/__init__.py
jrobinAV: The changes here are fully compatible with the linter rules on Core and Config. I believe we are all good.  

I made the same changes on a PR that includes other changes for release 2.3.
FYI, here is the [PR for release 2.3](https://github.com/Avaiga/taipy-gui/pull/694/files)

Issue #699, opened 2023-04-06
Tune the Stylekit


FabienLelaquais:
**Description**

A few things are not quite perfect:

- The margin has disappeared from the root page
- The 'svg-bg' class name makes little sense.
- The CSS class 'rows-similar' for tables is not documented
- The documentation for the CSS class 'align-item-stretch' of the part block is unclear.
- The names from the *bg-default*, *bg-default-light* and *bg-default-dark* are unfortunate, when looked at next to *color-background*, *color-background-light* and *color-background-dark*.
- The name for `--component-padding` must be changed. "Component", in Taipy, is "React Component".

** Proposal **
- margin: add a *root-margin* variable (defaults to 1em) that would set *--root-margin* used by `<body>`.
-  'svg-bg' -> *has-background*.
- 'rows-similar', 'align-item-stretch': check online working doc.
- *bg-default*, *bg-default-light* and *bg-default-dark* -> *bg-background*, *bg-background-light* and *bg-background-dark*
- `--component-padding` -> *--element-padding*

**Acceptance Criteria**
- [ ] Ensure any change is well documented

ghost:
> The margin has disappeared from the root page

This has been discussed just now with Vincent and Marine. This has been done to allow elements like header or sidebar to stick to the edges of the page. Putting a root margin is a bad practice (that’s why this is always reset on *every* page in CSS), and I intended to let the user choose the root margin by placing his app’s content into a part with a utility padding class like `p2`. I realize this feels to much more you guys, so I’d like to propose adding a Stylekit variable like `root_margin`, with a default value like `32px`, to easily override and tweak that value without going through custom CSS.
What do you think ?

ghost:
> The 'svg-bg' class name makes little sense.

I get this is not ideal but I had a hard time finding a better name. The intention was to use the original SVG background instead of the container's background.
Would `inner-bg` or `intrinsic-bg` sound better to you?

ghost:
> The CSS class 'rows-similar' for tables is not documented

Damn. My bad. That’s corrected in the [Google doc](https://docs.google.com/document/d/1VXMVPmZ8H3S-PSy05tr9Q6-9CR_Zit4yl1GzIHx1QJY/edit#).

ghost:
> The documentation for the CSS class 'align-item-stretch' of the part block is unclear.

[...]

PR #711, opened 2023-04-12
Stylekit: rename classes that deal with background color

ghost:
Address #699 ’s 5th point.

These are the *bg-default-* utility classes, renamed to *bg-background-* for consistency.


FabienLelaquais (approved):
👌

PR #712, opened 2023-04-13
Stylekit: subtract root-margin from the sidebar height

ghost:


FabienLelaquais (approved):
Probably just fine...

PR #713, opened 2023-04-13
Stylekit: rename chart modifier "svg-bg" class into "has-background" for clarity

ghost:
Address #699 second point.


FabienLelaquais (approved):
👍

Issue #719, opened 2023-04-14
BUG- Partial rendering error: Content not displayed


FlorianJacta:
**Description**
Partials are not working anymore. An error is shown instead of what should be inside the partial.

![image](https://user-images.githubusercontent.com/98709993/232068679-ff8fddeb-5fa8-413c-8398-c42bc42cfd75.png)

**How to reproduce**
```
from taipy import Gui

gui = Gui("<|part|partial={p}|>")
p = gui.add_partial("test")
gui.run()
```

**Expected behavior**
Partials should work like previous versions. I should have 'test' on the page.

**Runtime environment**
Taipy GUI: 2.2.0

PR #720, opened 2023-04-14
broken partial in part (#719)

FredLL-Avaiga:
from taipy.gui import Gui

gui = Gui("""
<|part|partial={p}|>

<|part|page=https://taipy.io|>

<|change partial|button|on_action={lambda s: p.update_content(s, 'from button')}|>
""")
p = gui.add_partial("test")
gui.run()


FabienLelaquais (approved):
I even think I get it

PR #721, opened 2023-04-14
#719 backport broken partial in part

FredLL-Avaiga:
v 2.2.1


FabienLelaquais (approved):
I don't really know what we do with the version for taipy-gui-dom.
As long as there's no strict equality test... I'm all good

FredLL-Avaiga:
> I don't really know what we do with the version for taipy-gui-dom. As long as there's no strict equality test... I'm all good

Just keep things in sync

PR #723, opened 2023-04-17
fix(stylekit): add max-width: 100% on taipy-table

ghost:
Prevent table width from exceeding parent width.


FredLL-Avaiga (approved):

FabienLelaquais (approved):
👍

PR #724, opened 2023-04-18
fix(stylekit): use a minHeight on inputs instead of a solid height.

ghost:
This avoids the issue of chips invisible on new lines when a multiple selection is done (like this example) and allows the input to grow as needed.
image


ghost:
@FredLL-Avaiga, @FabienLelaquais I’ll let you merge this if this is good for you.

FredLL-Avaiga (approved):

FabienLelaquais (approved):
👌

Issue #682, opened 2023-03-24
ssl_context doesn't work


livia0803:
**Description**

**How to reproduce**
A code fragment
```
from taipy import Gui

if __name__ == "__main__":
    Gui(page="Taipy Test").run(host="0.0.0.0", port=10002, ssl_context="adhoc")

```

**Expected behavior**
![image](https://user-images.githubusercontent.com/5668597/227482997-d9fc1de9-a228-493a-aff3-1a17f3db50b6.png)

**Screenshots**
<img width="1045" alt="image" src="https://user-images.githubusercontent.com/5668597/227480913-fe165acf-5b27-40eb-9058-3cfcf80706cf.png">

**Runtime environment**
Please specify relevant indications.
 - OS: Macos Ventura 13.2.1
 - Browser: Chrome 111.0.5563.64
and any other relevant information.

PR #729, opened 2023-04-20
Server Config (#734, #682)

dinhlongviolin1:
Solve Issue #734, #682

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -1682,15 +1682,13 @@ class Gui:
                         for n, e in lib.get_elements().items()
                         if isinstance(e, Element) and not e._is_server_only()
                     ]
-        stylekit = self._get_config("stylekit", _default_stylekit)
-        if stylekit:
+        if stylekit := self._get_config("stylekit", _default_stylekit):
             config["stylekit"] = {_to_camel_case(k): v for k, v in stylekit.items()}
         return config
 
     def __get_css_vars(self) -> str:
         css_vars = []
-        stylekit = self._get_config("stylekit", _default_stylekit)
-        if stylekit:
+        if stylekit := self._get_config("stylekit", _default_stylekit):
             for k, v in stylekit.items():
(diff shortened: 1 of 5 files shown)


src/taipy/gui/gui.py
```diff
@@ -1704,6 +1702,7 @@ def __init_server(self):
                 path_mapping=self._path_mapping,
                 flask=self._flask,
                 async_mode=app_config["async_mode"],
+                server_config=app_config["server_config"],
```
FredLL-Avaiga: should we use app_config.get("server_config") here ?

src/taipy/gui/gui.py
```diff
@@ -1715,6 +1714,7 @@ def __init_server(self):
...
```
FredLL-Avaiga: and here

FredLL-Avaiga (commented):
Looks like szerver_config can be None ?

src/taipy/gui/server.py
```diff
@@ -50,26 +51,45 @@ def __init__(
...
         gui: Gui,
         flask: t.Optional[Flask] = None,
-        path_mapping: t.Optional[dict] = {},
+        path_mapping: t.Optional[dict] = None,
         async_mode: t.Optional[str] = None,
+        server_config: t.Optional[ServerConfig] = None,
     ):
         self._gui = gui
+        server_config = server_config or {}
         self._flask = Flask("Taipy") if flask is None else flask
```
FabienLelaquais: Do we want to let users customize the name of the application package here?
Adding a slot in ServerConfig..?

FabienLelaquais (approved):
Yes I like that a lot!
Well done!

src/taipy/gui/server.py
dinhlongviolin1: I have made it so you configure all flask init parameter

FredLL-Avaiga (approved):

PR #730, opened 2023-04-21
Feature/taipy#126 group taipy cli args to subcommands

trgiangdo:
In this PR:
- Change the behavior (or a bug) where the arguments loaded from the toml file will override CLI arguments. It should be the other way around.
- Add loading args from TOML and CLI test case.

### DIFF
--- a/src/taipy/gui/_gui_cli.py
+++ b/src/taipy/gui/_gui_cli.py
@@ -9,15 +9,26 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
-from taipy.config.common._argparser import _Argparser
+from argparse import Namespace, ArgumentError
+from taipy.config._cli._argparser import _Argparser
 
 
 class _GuiCLI:
     """Command-line interface of GUI."""
+    _DEFAULT_ARGS = {
+        "port": "",
+        "host": "",
+        "ngrok_token": "",
+        "webapp_path": "",
+        "debug": False,
+        "no_debug": False,
+        "use_reloader": False,
+        "no_reloader": False,
+    }
 
     @classmethod
-    def _create_parser(cls):
-        gui_parser = _Argparser._add_groupparser("GUI", "Optional arguments for GUI service")
+    def create_parser(cls):
+        gui_parser = _Argparser._add_subparser("taipy", help="Run application with Taipy arguments.")
 
         gui_parser.add_argument("-P", "--port", nargs="?", default="", const="", help="Specify server port")
         gui_parser.add_argument("-H", "--host", nargs="?", default="", const="", help="Specify server host")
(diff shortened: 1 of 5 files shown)


tests/taipy/gui/cli/test_cli.py
```diff
@@ -0,0 +1,155 @@
...
+
+    # Override Config.configure_gui() arguments by loading a TOML file with [gui] section
+    toml_config = NamedTemporaryFile(
+        content="""
+[TAIPY]
+
+[gui]
+host = "my_3rd_host"
+port = 5003
+use_reloader = "true:bool"
```
FabienLelaquais, line 140: I don't understand what that is not simply
`use_reloader = true`

That would be very valid TOML...

src/taipy/gui/_gui_cli.py
```diff
@@ -9,15 +9,26 @@
...
+        "no_debug": False,
+        "use_reloader": False,
+        "no_reloader": False,
+    }
 
     @classmethod
-    def _create_parser(cls):
-        gui_parser = _Argparser._add_groupparser("GUI", "Optional arguments for GUI service")
+    def create_parser(cls):
+        gui_parser = _CLI._add_subparser("taipy", help="Run application with Taipy arguments.")
```
FabienLelaquais: This I think breaks the legacy parsing mechanism.
Users would for example use `python script.py -debug` and get what they expect.
Now it seems that forces them to use `python script.py taipy -debug`.

FabienLelaquais (changes requested):
Please make sure my worry is addressed.

src/taipy/gui/_gui_cli.py
FabienLelaquais: What I'm saying is that we need to rethink this thing and try to allow not to use the `taipy` option

tests/taipy/gui/cli/test_cli.py
trgiangdo: In taipy-config, we are using the `:bool` to separate a boolean from a string.

This is just how it was implemented.

FredLL-Avaiga (approved):

src/taipy/gui/_gui_cli.py
```diff
@@ -9,15 +9,15 @@
...
 
 
 class _GuiCLI:
     """Command-line interface of GUI."""
 
     @classmethod
-    def _create_parser(cls):
-        gui_parser = _Argparser._add_groupparser("GUI", "Optional arguments for GUI service")
+    def create_parser(cls):
+        gui_parser = _CLI._add_groupparser("Taipy GUI", "Optional arguments for Taipy GUI service")
```
FabienLelaquais, line 20: What's the point of the 'Taipy ' prefix?

FabienLelaquais (approved):
Looks good to me!

src/taipy/gui/_gui_cli.py
trgiangdo: Just that when we show the help message, it is clear that these options are for Taipy GUI

Issue #683, opened 2023-03-24
Unable to prefix all requests


livia0803:
**Description**
I'd like to set all the url start with `127.0.0.1:5102/abc/`, but I can't find a way to prefix all requests.

I have tried several methods
-  run `Gui` with `Gui(pages=pages).run(host="0.0.0.0", port=5102)` and set pages with `abc`.
```python 

from taipy import Gui

pages = {"abc/page1": "# page1", "abc/page2": "# page2"}

Gui(pages=pages).run(host="0.0.0.0", port=5102)

```

-  run `Gui` with `Gui(flask=app).run(host="0.0.0.0", port=5102)` and set app.url_rule_class = Custom_Rule.Here is the code  
```python


from taipy import Gui
from flask import Flask
from werkzeug.routing import Rule

app = Flask(__name__)

APP_ROOT = "abc"


class Custom_Rule(Rule):
	def __init__(self, string, *args, **kwargs):
		# check endswith '/'
		if APP_ROOT.endswith('/'):
			prefix_without_end_slash = APP_ROOT.rstrip('/')
		else:
			prefix_without_end_slash = APP_ROOT
		# check startswith '/'
		if APP_ROOT.startswith('/'):
			prefix = prefix_without_end_slash
		else:
			prefix = '/' + prefix_without_end_slash
		super(Custom_Rule, self).__init__(prefix + string, *args, **kwargs)


app.url_rule_class = Custom_Rule


@app.route("/")
def root():
	return {"root": True}


@app.route("/home")
def home():
	return {"home": True}


Gui(flask=app).run(host="0.0.0.0", port=5102)

```
**Both methods above only apply to customized requests.**   Such as: `abc/page1`, `abc/home`

**Expected behavior**
set all request start with `127.0.0.1:5102/abc/`. Applies to all request types:document, script, javascript, xhr, and so on.

**Screenshots**
<img width="686" alt="image" src="https://user-images.githubusercontent.com/5668597/227495219-1e551939-1cd7-4712-adae-b3c39105edc8.png">
**Runtime environment**
Please specify relevant indications.
 - OS: [e.g. Linux, Windows] and version
 - Browser: [e.g. Chrome, Edge, Safari] and version
and any other relevant information.

PR #733, opened 2023-04-25
Client URL Prefix (#683, #735)

dinhlongviolin1:
Solve Issue #683 #735


gui/src/utils/index.ts
```diff
@@ -161,3 +161,7 @@ export const getInitials = (value: string, max = 2): string =>
         .toUpperCase();
 
 export const TIMEZONE_CLIENT = Intl.DateTimeFormat().resolvedOptions().timeZone;
+
+export const getBaseURL = (): string => {
+    return document.querySelector("base")?.getAttribute("href") || "/";
```
FredLL-Avaiga: could have used a window.taipyConfig value instead of parsing the HTML at each render of a component ?

src/taipy/gui/config.py
```diff
@@ -292,3 +295,13 @@ def _resolve_stylekit(self):
...
                 app_config["stylekit"]["root_margin"] = str(app_config["margin"])
             app_config["margin"] = None
+
+    def _resolve_url_prefix(self):
+        app_config = self.config
+        client_url_prefix = app_config.get("client_url_prefix")
+        if client_url_prefix is not None:
+            app_config[
+                "client_url_prefix"
+            ] = f"{'' if client_url_prefix.startswith('/') else '/'}{client_url_prefix}{'' if client_url_prefix.endswith('/') else '/'}"
```
FredLL-Avaiga: Can we have client_url_prefix with several levels ? ie /ff/dd/zz/ ?

FredLL-Avaiga (approved):
a bit worried about "document.querySelector"

gui/src/utils/index.ts
FabienLelaquais: Is it really parsing though?
The DOM is loaded in the browser... it's just a query and I bet that this would be pretty fast.
Selecting &lt;head&gt; first would garantee that I guess.

FredLL-Avaiga: slower than window.taipyConfig.baseUrl I guess ...

src/taipy/gui/config.py
dinhlongviolin1: You can have as many levels as you want to :D

gui/src/utils/index.ts
dinhlongviolin1: @FredLL-Avaiga You are right, window.taipyConfig can also work. The reason why I used dom parsing is to prevent redundancy within the html file

src/taipy/gui/_default_config.py
```diff
@@ -45,6 +45,7 @@
     "async_mode": "gevent",
     "change_delay": None,
     "chart_dark_template": None,
+    "client_url_prefix": "/",
```
FabienLelaquais: Why the "client_" prefix?
What would a use case where "url_prefix" has to be used, NOT on the client?

FabienLelaquais (approved):
Neat!!

gui/src/utils/index.ts
FredLL-Avaiga: A few bytes, once in a while :-)

src/taipy/gui/_default_config.py
FabienLelaquais: We decided to go for "base_url"

src/taipy/gui/config.py
```diff
@@ -102,6 +103,7 @@
         "async_mode": str,
         "change_delay": t.Optional[int],
         "chart_dark_template": t.Optional[t.Dict[str, t.Any]],
+        "base_url": t.Optional[str],
```
FredLL-Avaiga, line 106: Is this optional?

src/taipy/gui/config.py
```diff
@@ -311,6 +314,14 @@ def _resolve_stylekit(self):
                 app_config["stylekit"]["root_margin"] = str(app_config["margin"])
             app_config["margin"] = None
 
+    def _resolve_url_prefix(self):
+        app_config = self.config
+        base_url = app_config.get("base_url")
+        if base_url is not None:
```
FredLL-Avaiga: What if baseUrl is set at ""?

FredLL-Avaiga (approved):
😮

.github/workflows/prefix-test.yml
```diff
@@ -0,0 +1,57 @@
...
+        working-directory: ./gui
+      - name: install libmagic on macos
+        if: matrix.os == 'macos-latest'
+        run: brew install libmagic
+
+      - name: install pipenv and install dependencies
+        run: |
+          pip install pipenv
+          pipenv install --dev
+          pipenv run pip install jsonschema==4.17.3
```
FabienLelaquais, line 53: Isn't this going to haunt us for a while?

gui/src/components/Taipy/NavBar.tsx
```diff
@@ -56,7 +56,7 @@ const NavBar = (props: LovProps) => {
     const linkChange = useCallback(
         (evt: React.SyntheticEvent, val: string) => {
             if (Object.keys(state.locations || {}).some((route) => val === route)) {
-                navigate(val);
+                navigate(val.substring(1));
```
FabienLelaquais, line 59: Is the target url always absolute?

FabienLelaquais (approved):
Really well done!

.github/workflows/prefix-test.yml
dinhlongviolin1: this should be removed once jsonschema release a new version after the dev one

gui/src/components/Taipy/NavBar.tsx
dinhlongviolin1: it should be absolute so the baseurl can take proper affect. i might be wrong but we can always fix this in the future :D

src/taipy/gui/config.py
dinhlongviolin1: You got a point, thanks!

src/taipy/gui/config.py
dinhlongviolin1: yes, it could be optional

Issue #741, opened 2023-05-04
BUG-allow_unsafe_werkzeug in Notebook


FredLL-Avaiga:
**Description**
From Socketio 5.3, the allow_unsafe_werkzeug=True is mandatory to run flask without a production ready webserver
In notebook context, use_reloader interacts with allows_unsafe_werkzeug

**How to reproduce**
run any taipy app with socketio 5.3 

**Expected behavior**
ignore use_reloader in notebook

**Runtime environment**
- socketio 5.3+
- taipy 2.2+

FredLL-Avaiga:
Looks like use_reloader is a culprit here 
It forces the async_mode to threading which is not compatible with allow_unsafe_werkzeug
We need to prevent use_reloader in notebook.

Issue #619, opened 2023-02-22
Better performance on re-running on existing port for Notebooks


FlorianJacta:
**Description**
After running a first app on a Notebook (`gui.run(...)`), re-running the app (`gui.run(...)`) takes a long time (more than 15 seconds). The goal would be to shorten this time to the maximum.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FlorianJacta:
It will load fast when we hit a lot of time on F5 to refresh the page. But if we hit just once F5 to refresh, it will always try to refresh without success (the page tells me it is trying to refresh).

Process:
- Write a Taipy Gui code in Jupyter Notebook or VS Code notebook...
- Run the cell that runs the GUI
- It opens a tab with my application (it is fast)
- I change my Markdown and rerun the cell where the GUI was run. My cell is executed with these messages:
```
Gui server has been stopped.
 * Server starting on http://127.0.0.1:5000/
```
- Go back to my tab and refresh with F5
- It will never refresh

- If I hit F5 a lot of times, it will refresh in a short amount of time

![image](https://user-images.githubusercontent.com/98709993/229556766-1ac32a23-fcaa-42c1-8fa5-42ec9dfc9dcf.png)

PR #742, opened 2023-05-05
Notebook proxy (#619, #741)

dinhlongviolin1:
Adresses #741, #619.

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -23,6 +23,7 @@ simple-websocket = "==0.9"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
 tzlocal = "==3.0"
 gitignore-parser = "==0.1.1"
+twisted = "==22.10.0"
 
 [dev-packages]
 black = "*"
--- a/src/taipy/gui/config.py
+++ b/src/taipy/gui/config.py
@@ -43,6 +43,7 @@ ConfigParameter = t.Literal[
     "light_theme",
     "margin",
     "ngrok_token",
+    "notebook_proxy",
     "notification_duration",
     "propagate",
     "run_browser",
@@ -112,6 +113,7 @@ Config = t.TypedDict(
         "light_theme": t.Optional[t.Dict[str, t.Any]],
         "margin": t.Optional[str],
         "ngrok_token": str,
+        "notebook_proxy": bool,
         "notification_duration": int,
         "propagate": bool,
         "run_browser": bool,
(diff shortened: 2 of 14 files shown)


Pipfile
```diff
@@ -23,6 +23,7 @@ simple-websocket = "==0.9"
 taipy-config = {git = "https://git@github.com/avaiga/taipy-config.git@develop"}
 tzlocal = "==3.0"
 gitignore-parser = "==0.1.1"
+twisted = "==22.10.0"
```
FredLL-Avaiga, line 26: Should this be optional ?

src/taipy/gui/config.py
```diff
@@ -271,6 +273,10 @@ def resolve(self):
                 logger, "'use_reloader' parameter will not be used when 'ngrok_token' parameter is available"
             )
 
+        if app_config["use_reloader"] and _is_in_notebook():
```
FredLL-Avaiga, line 276: related to #741

FredLL-Avaiga (approved):
I think we should make the twisted package optional

src/taipy/gui/config.py
FabienLelaquais: Makes sense.

src/taipy/gui/gui.py
```diff
@@ -1994,8 +1996,23 @@ def run(
             flask_log=app_config["flask_log"],
             run_in_thread=app_config["run_in_thread"],
             allow_unsafe_werkzeug=app_config["allow_unsafe_werkzeug"],
+            notebook_proxy=app_config["notebook_proxy"],
         )
 
+    def reload(self):
```
FabienLelaquais: Are you saying that, instead of stop()/run(), users should call 'reload()' when using Notebooks?

Pipfile
FabienLelaquais: I guess is could: only makes sense when running within Notebooks...
But then how do we address the versioning problem?

FabienLelaquais (approved):
Minor (?) comments

Pipfile
dinhlongviolin1: We should keep this as required installation in pipfile since it directly affects the notebook side of things

src/taipy/gui/gui.py
dinhlongviolin1: that is what is on my mind. if you dont pass anything into the run function then it doesnt make any difference. But if you do, it will preserve the config and reloads everything

Issue #737, opened 2023-05-03
State: refresh method


FredLL-Avaiga:
**What would that feature address**
A nicer way of asking for a variable refresh instead of state.var = state.var

***Description of the ideal solution***
state.refresh("var")
As discussed with @vhfgosselin 

***Caveats***
None

***Other options***
 state.var = state.var

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #744, opened 2023-05-05
state.refresh (#737)

FredLL-Avaiga:
Addresses #737.


FabienLelaquais (approved):
But why...

FredLL-Avaiga:
> But why...

That's a good point. Should we ?
Vincent thought that it'd be nicer than state.var = state.var

Issue #595, opened 2023-01-31
BUG- variable from a page not available in on_navigate


FlorianJacta:
**Description**
An error is raised when trying to change the value of a variable. This variable is in the Markdown of an imported page. The value is also not changed.

`UserWarning: Variable 'number' is not available in either the '__main__' or '__main__' modules.`

**How to reproduce**
Here the goal is to change the _number_ value when we navigate to the _data_ page.

In _home.py_:
```python
from taipy.gui import Markdown

home_md = Markdown("Nothing")
```

In _data.py_:
```python
from taipy.gui import Markdown

number = 1

data_md = Markdown("<|{number}|>")
```

In _main.py_:
```python
from data import data_md
from home import home_md

from taipy.gui import Markdown, Gui

pages = {
    "home": home_md,
    "data": data_md,
    "/": Markdown("<|navbar|>")
}

def on_navigate(state, page_name: str) -> str:
    if page_name == "data":
        state.number = 0
    return page_name

Gui(pages=pages).run(port=5050)
```

**Expected behavior**
_number_ should be put at 0 when changing the page. No error should be raised.

**Runtime environment**
Please specify relevant indications.
 - Taipy GUI: 2.1.0

arcanaxion:
I see that this behaviour is probably intentional since the `number` variable may be in more than one of the modules. Hence, related to this, is there an alternative way that the user can specify the module in which `number` is to be modified?

FredLL-Avaiga:
I guess it would work if you import number in the main module ...

FlorianJacta:
Yes, but this could be better. Here is just a simple example that is easy to fix. On a complex app, it becomes a nightmare to import or not elements from other pages to the main. It occurs not only for _on_navigate_ but also for _on_init_ and maybe other Taipy GUI basic functions. I have seen it firsthand working on the new demo to come :

https://github.com/Avaiga/demo-qsr-sales-forecasting-dashboard

I think the ideal solution would be something like this:


In _home.py_:
```python
from taipy.gui import Markdown

home_md = Markdown("Nothing")
```

In _data.py_:
```python
from taipy.gui import Markdown

number = 1

data_md = Markdown("<|{number}|>")

def on_navigate(state, page_name: str):
   state.number = 1
```

In _main.py_:
```python
import data
import home

from taipy.gui import Markdown, Gui

pages = {
    "home": home.home_md,
    "data": data.data_md,
    "/": Markdown("<|navbar|>")
}

[...]

PR #745, opened 2023-05-15
allow state to access correct page scopes at runtime (#595, #751)

dinhlongviolin1:
Solve issue #595, #751


FredLL-Avaiga (approved):
Magic 🪄

FabienLelaquais (approved):
Amazing. I hope we won't be hurt too much when Python decides to change its frame management...
Well done!!

Issue #755, opened 2023-05-22
BUG-Table.width default value is not adequate for stylekit


FredLL-Avaiga:
**Description**
Table.width default value is not adequate for stylekit.

**How to reproduce**
A table with no width set

**Expected behavior**
Width should default to 100%

PR #756, opened 2023-05-22
Table width: set default to 100% (#755)

FredLL-Avaiga:
Addresses #755.

+ set default css filename to taipy.css when not in a python file (cf. taipy-doc/#403)


src/taipy/gui/gui.py
```diff
@@ -280,7 +280,9 @@ def __init__(
         if css_file is None:
             script_file = pathlib.Path(self.__frame.f_code.co_filename or ".").resolve()
             if script_file.with_suffix(".css").exists():
-                css_file = f'{script_file.stem or "Taipy"}.css'
+                css_file = f"{script_file.stem}.css"
+            elif script_file.is_dir() and (script_file / "taipy.css").exists():
```
FabienLelaquais, line 284: Will you let me know how to document this case?

FabienLelaquais (approved):
I think that's neat!

src/taipy/gui/gui.py
FredLL-Avaiga: I guess that when in a notebook script_file => "." (need to be confirmed)
In this case (and in every case where co_filename is None), we look for a taipy.css file in the current directory...
Not sure what current directory means in a notebook but we can find out ...

dinhlongviolin1 (approved):

PR #762, opened 2023-05-31
fix(selector): reposition label on top of the selector in list view

ghost:
fix(selector): reposition label on top of the selector in list view (different from dropdown view).

Before (issue raised by Florian J): 
image

After:
image


FredLL-Avaiga (approved):
no easy to find the 2 lines change

PR #765, opened 2023-06-02
Better document for the table control.

FabienLelaquais:
Addresses [doc#441](https://github.com/Avaiga/taipy-doc/issues/441).


gui/doc/table.md
```diff
@@ -38,155 +58,296 @@ tables:
...
+the name of a style, defined in CSS.
 
-The _group_by[column_name]_ property, when set to _True_ indicates that the column _column_name_ can be
-aggregated.
+Here is how a row styling function can be defined:
 
-The function provided in the _apply[column_name]_ property indicates how to perform this aggregation.
-The value of this property, which is a string, can be:
+```py
+def even_odd_style(_1, index, _2):
```
FredLL-Avaiga, line 230: I think you can omit the 3rd parameter (_2)

FredLL-Avaiga (approved):

gui/doc/table.md
FabienLelaquais: I think you are right. It is a bit clearer as such in my opinion, with respect to the documentation of the function signature.

Issue #766, opened 2023-06-02
Styling of component in panes


ghost:
**Description**
Hi everyone,
it seems that the styling of element in panes is not fully taken into account.

**How to reproduce**
Simple app :
```
from taipy.gui import Gui, notify
import pandas as pd

#list of kpi (key / value)
tp_kpis = pd.DataFrame(
    {
        "name":["A", "B", "C"],
        "value":["1", "2", "3"]
    }
)

show_kpi = False

def display_kpi(state):
    state.show_kpi = True

main_page = """
<|toggle|theme|>
###Kpis
<|Kpis|button|on_action=display_kpi|>
<|{show_kpi}|pane|anchor=bottom|width=560px|
<|card|
<|{tp_kpis}|table|columns=name; value|show_all=True|width=500px|height=700px|>
|>
|>

<|card|
<|{tp_kpis}|table|columns=name; value|show_all=True|width=500px|height=700px|>
|>
"""
#Run Taipy GUI
gui = Gui(main_page).run(dark_mode=True)
```

**Expected behavior**
Have the proper styling in the pane

**Screenshots**

![image](https://github.com/Avaiga/taipy-gui/assets/102367972/cb1d39bc-2a57-42b1-8878-11972b2593c8)

**Runtime environment**
Please specify relevant indications.
 - OS: Ubuntu 22.04.2 LTS
 - Browser: Chromium

FlorianJacta:
Thank you for your issue! It seems that visual elements or Stylekit components (like tables or card) behaves wrongly in the pane control. We will work on it!

I created a simpler code to reproduce this behavior:

```python
from taipy.gui import Gui

data = {
        "name":["A", "B", "C"],
        "value":["1", "2", "3"]
    }


main_page = """
<|toggle|theme|>

<|{True}|pane|
<|{data}|table|>

<|card|
Here is a **card**{.color-primary}
|>
|>
"""

Gui(main_page).run(dark_mode=False)
```

Screenshot:

![image](https://github.com/Avaiga/taipy-gui/assets/98709993/40964eb6-dae2-410b-b7f0-4562fda07183)

PR #767, opened 2023-06-05
fix(pane): add mode class on the pane

ghost:
Address #766 reported by a user.

## Problem
As the Drawer is out of the taipy-dark/light class div, dark mode has no effect on it, especially with the Stylekit that relies on that class name.
pane

## Solution
Add the mode-aware class name on the pane to enable color mode capabilities on that component without moving it into the root div.

### DIFF
--- a/gui/src/components/Taipy/Pane.tsx
+++ b/gui/src/components/Taipy/Pane.tsx
@@ -11,13 +11,7 @@
  * specific language governing permissions and limitations under the License.
  */
 
-import React, {
-  ReactNode,
-  useCallback,
-  useEffect,
-  useMemo,
-  useState,
-} from "react";
+import React, { ReactNode, useCallback, useEffect, useMemo, useState } from "react";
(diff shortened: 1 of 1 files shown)


FredLL-Avaiga (commented):
Global reformat makes it very uneasy to spot the "real" changes :-(
Another solution would be to change the theme class name at the root level (HTML) ?

ghost:
@FredLL-Avaiga Yes sorry I saw that too and updated that right before you posted your comment :D
Forgot to disable my code formatter.

Yes, changing that at the `<html>` level would be better indeed. But I didn’t see how (and where) to do that in the project structure.

FredLL-Avaiga (commented):
I proposed that we drop this PR and we change the current implementation to set the theme class to the body.
What do you think @JaufreyTpy ?

ghost:
@FredLL-Avaiga Fine by me. So I let you change the implementation ?

Issue #690, opened 2023-03-30
Enhance organization of complex applications


FlorianJacta:
## Faced issue

Building complex applications requires manipulating a lot of Markdown and code surrounding it to create the logic behind it. The goal is to find an efficient way to organize an application between several pages and several sections of your pages. A page with a lot of functionality will be harder to understand and maintain if represented in a single Markdown file and with a long Python code associated with it.

This issue aims to have ways or best practices for creating complex applications in a structured and intuitive manner. This would help further improve the readability, maintainability, and navigation of applications, making the development process even more seamless.

## Desired outcome:

- Clear and intuitive guidance for organizing applications into pages and sections
- Enhanced readability and maintainability of code and markdown
- Improved navigation within the application structure

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

dinhlongviolin1:
#754, #751 is related

PR #769, opened 2023-06-05
Class Modules (#690)

dinhlongviolin1:
Solve Issue #690 

To be reviewed after 2.3 is released!!!

```python
from src.taipy.gui import Markdown, Gui, ClassModule


class TestModule(ClassModule):
    def __init__(self):
        self.x = 10
        self.y = 20
        self.page = Markdown(
            """
x = <|{x}|>

<|Set x to 30|button|on_action=set_x|>
"""
        )
    def set_x(self):
        print("hey")
        self.x = 30

    def on_change(self):
        pass


gui = Gui()
gui.add_page("test", TestModule)
gui.run(run_browser=False)
```


src/taipy/gui/page.py
```diff
@@ -85,10 +87,27 @@ def _get_content_detail(self, gui) -> str:
...
+            self._class_locals
+            if self._class_locals != {}
+            else None
+            if self._frame is None
+            else _filter_locals(self._frame.f_locals)
+        )
+
+    def _set_class_module(self, name: str, locals_: t.Dict[str, t.Any]) -> None:
+        self._class_module_name = name
+        self._class_locals = locals_
```
FredLL-Avaiga: What happens if class has no other member than page?

FredLL-Avaiga (approved):
simple & effective 👍

src/taipy/gui/page.py
dinhlongviolin1: then it should be an empty locals, as it should be

src/taipy/gui/__init__.py
```diff
@@ -70,6 +70,7 @@
 
 from importlib.util import find_spec
 
+from .class_module import ClassModule
```
FabienLelaquais, line 74: My proposal - to be discussed
`from .base_page import BasePage`

FabienLelaquais (approved):
I totally love it.
So simple! So neat!

dinhlongviolin1:
This one is obsolete, replaced by PR #816

PR #770, opened 2023-06-05
set theme class to body (#766)

FredLL-Avaiga:
Addresses #766.

So that every component in the page is impacted

### DIFF
--- a/gui/src/components/Router.tsx
+++ b/gui/src/components/Router.tsx
@@ -92,6 +92,16 @@ const Router = () => {
         initializeWebSocket(state.socket, dispatch);
     }, [state.socket]);
 
+    useEffect(() => {
+        const classes = [themeClass];
+        document.body.classList.forEach((cls) => {
+            if (!cls.startsWith("taipy-")) {
+                classes.push(cls);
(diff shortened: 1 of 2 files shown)


gui/src/components/Router.tsx
```diff
@@ -92,6 +92,16 @@ const Router = () => {
         initializeWebSocket(state.socket, dispatch);
     }, [state.socket]);
 
+    useEffect(() => {
+        const classes = [themeClass];
+        document.body.classList.forEach((cls) => {
+            if (!cls.startsWith("taipy-")) {
```
FabienLelaquais, line 98: That would remove all taipy- classes. Isn't that a bit dangerous?

FabienLelaquais (approved):
👌

gui/src/components/Router.tsx
FredLL-Avaiga: I know: there are none except now for theme and taipy-gui should manage this so ...

dinhlongviolin1 (approved):

ghost (approved):

Issue #752, opened 2023-05-19
BUG-jinja2.exceptions.TemplateNotFound: index.html when starting gui web server


arcanaxion:
**Description**
Installing the current develop branch of taipy-gui, starting a simple application then opening the webpage shows:

![image](https://github.com/Avaiga/taipy-gui/assets/48582724/76d5968f-af9f-495d-b81e-c71d98b1a6f1)

Terminal shows `jinja2.exceptions.TemplateNotFound: index.html` (full traceback below).

**How to reproduce**
```python
# main.py
from taipy.gui import Gui

Gui(page="Hello, world").run()
```

```bash
> pyenv shell 3.10
> python -m venv .venv
> source .venv/bin/activate
> pip install "git+https://git@github.com/avaiga/taipy-gui"
> python main.py
```

Current taipy-gui commit: 58a25ca61056f1b73f158329bf8a068af063ef75

**Expected behavior**
Hello world should display on page as normal.

Traceback:

```
basic-taipy-app  | [2023-05-19 01:25:08][Taipy][INFO]  * Server starting on http://localhost:5000
basic-taipy-app  | [2023-05-19 01:25:39][Taipy][ERROR] Exception on / [GET]
basic-taipy-app  | Traceback (most recent call last):
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 2529, in wsgi_app
basic-taipy-app  |     response = self.full_dispatch_request()
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1825, in full_dispatch_request
basic-taipy-app  |     rv = self.handle_user_exception(e)
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/flask_cors/extension.py", line 165, in wrapped_function
basic-taipy-app  |     return cors_after_request(app.make_response(f(*args, **kwargs)))
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1823, in full_dispatch_request
basic-taipy-app  |     rv = self.dispatch_request()
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/flask/app.py", line 1799, in dispatch_request
basic-taipy-app  |     return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
basic-taipy-app  |   File "/usr/local/lib/python3.10/site-packages/taipy/gui/server.py", line 150, in my_index
[...]

PR #771, opened 2023-06-05
#752 Hint on can't load index template error message

FredLL-Avaiga:
@FabienLelaquais If you have a better message ...

### DIFF
--- a/src/taipy/gui/server.py
+++ b/src/taipy/gui/server.py
@@ -147,19 +147,25 @@ class _Server:
         @taipy_bp.route("/<path:path>")
         def my_index(path):
             if path == "" or path == "index.html" or "." not in path:
-                return render_template(
-                    "index.html",
-                    title=title,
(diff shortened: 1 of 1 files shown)


src/taipy/gui/server.py
```diff
@@ -147,19 +147,25 @@ def _get_default_blueprint(
...
+                        config=client_config,
+                        scripts=scripts,
+                        styles=styles,
+                        version=version,
+                        css_vars=css_vars,
+                        base_url=base_url,
+                    )
+                except Exception:
+                    raise RuntimeError(
+                        "Something is wrong with the taipy-gui front-end installation. Check that the js bundle has been properly built (is NodeJs installed?)."
```
FabienLelaquais: I think the proper spelling is "Node.js"

FabienLelaquais (approved):
👍

Issue #785, opened 2023-06-10
BUG-part iframe size


FredLL-Avaiga:
**Description**
part iframe height/width is not managed

**How to reproduce**
A code fragment
```
<|part|page=http:docs.taipy.io|>
```

**Expected behavior**
Be able to set the height of the iframe inside the part

**Runtime environment**
Taipy-gui 2.2

Issue #784, opened 2023-06-09
Need to set height and width of `iframe` with `on_user_content()`


Dr-Irv:
* Taipy version: 2.2
* Python version:  3.10
* Operating System:  Windows

### Description

Trying to use the new `on_user_content()` feature as described here:  https://github.com/Avaiga/taipy-gui/issues/669#issuecomment-1476339616

The problem here is that the `iframe` that the content is set inside has a default height and width that I want to override.  In the example that I showed, I used ` html_string = f'<iframe src="/{filename}" height="{height}" width="100%" style="border:none;" scrolling="no"></iframe>'` .  I need to be able to set the `height` and `width` (and other properties as well, ideally)

@FredLL-Avaiga as implemented, the feature isn't that useful because we need control over the size of the generated `iframe` !!

FredLL-Avaiga:
We came across that same problem today :-(
The workaround is to use css for now.
But we'll add something about that!

Dr-Irv:
I hope the CSS workaround can be dynamic, i.e., where the height of the iframe is based on the embedded content.

What I am doing now is embedding a plotly `timeline`, and setting the height of the plot based on the number of "rows" of the Gantt chart.  Then I can use that value to get the height of the iframe .

FredLL-Avaiga:
Sadly it won't be be (css dynamic)
But we might be able to include a dynamic height property to the part while
setting the iframe's width and height to 100% in the upcoming 2.3 release...

FredLL-Avaiga:
cf. #785
https://github.com/Avaiga/taipy-gui/pull/786

FredLL-Avaiga:
@Dr-Irv This is available in taipy-gui dev release 2.3.0.dev5
https://pypi.org/project/taipy-gui/#history

Dr-Irv:
Thanks.  I installed that and now my Gantt chart displays.

PR #786, opened 2023-06-10
part height is dynamic (#785)

FredLL-Avaiga:
Addresses #784 and [doc#458](https://github.com/Avaiga/taipy-doc/issues/458).

iframe is 100% in size

```
from taipy.gui import Gui

height = "50vh"

page = """
#Part with iframe size

<|{height}|input|>

<|1 1|layout|

<|part|page=http://docs.taipy.io|>

<|part|page=http://taipy.io|height={height}|>
|>
"""

Gui(page).run()

```

### DIFF
--- a/gui/src/components/Taipy/Part.tsx
+++ b/gui/src/components/Taipy/Part.tsx
@@ -27,7 +27,14 @@ interface PartProps extends TaipyBaseProps {
     children?: ReactNode;
     defaultPartial?: boolean;
     partial?: boolean;
+    height?: string;
+    defaultHeight?: string;
 }
+
+const IframeStyle = {
+    width: "100%",
+    height: "100%",
+};
 
 const Part = (props: PartProps) => {
     const { id, children, partial, defaultPartial } = props;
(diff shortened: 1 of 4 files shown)


Dr-Irv:
Ideally, you could specify a class for the `iframe` that would allow you to set other `iframe` properties that are more static and could be different based on which iframe was being used.  (e.g., `style="border:none;overflow:hidden;"`

Or would the `class` that is on the enclosing `part` do the job?

FredLL-Avaiga:
I think we'll let the user define those style

FabienLelaquais (approved):
That's exactly what I was coding (I added a useless width prop to part).
Except yours probably works...

Dr-Irv:
> I think we'll let the user define those style

So you're saying that if I do `<|part|class_name=foo|page=mypage.html!>` and then I use `on_user_content()` linked to `mypage.html`, then the CSS for the class `foo` will be applied to the inner `iframe` that retrieves `mypage.html` ?

FredLL-Avaiga:
You'll have to declare something like
.foo>iframe {
  border:0;
}
In you css file

dinhlongviolin1 (approved):

PR #795, opened 2023-06-14
fix typing of add_pages()

Dr-Irv:
If you do
```python
pages = {"/" : "### Page1", "/foo" : "### Page 2"}
Gui=gui()
gui.add_pages(pages)
```
then pyright (in VS Code) complains that `pages` is incorrect type. 

That's because you have the type as `dict[Union[str, Page]]` but `dict` is invariant, while `Mapping` is covariant.

So this fixes that issue.

For the pre-commit to work, I also had to fix something in `types.py` because `mypy` was complaining.

Also fixed the pre-commit that grabs flake8 from gitlab, but it is on github.  And since I added overloads, had to add an error code to the `.flake8` configuration.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -2,7 +2,7 @@
 # required by black, https://github.com/psf/black/blob/master/.flake8
 max-line-length = 120
 max-complexity = 18
-ignore = E203, E266, E501, E722, W503, F403, F401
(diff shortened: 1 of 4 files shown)


dinhlongviolin1:
@Dr-Irv Thanks for the PR. LGTM!

PR #799, opened 2023-06-18
Table's data and editable[] properties are dynamic.

FabienLelaquais:
Small doc issue.

### DIFF
--- a/src/taipy/gui/viselements.json
+++ b/src/taipy/gui/viselements.json
@@ -970,7 +970,7 @@
           },
           {
             "name": "editable[<i>column_name</i>]",
-            "type": "dynamic(bool)",
+            "type": "bool",
             "default_value": "editable",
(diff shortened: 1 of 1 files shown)


src/taipy/gui/viselements.json
```diff
@@ -970,7 +970,7 @@
           },
           {
             "name": "editable[<i>column_name</i>]",
-            "type": "bool",
+            "type": "dynamic(bool)",
```
FredLL-Avaiga: I reread the code and this is not dynamic :-(
I think the idea was that you define which columns can be edited and they become editable when the global (dynamic) editable is set to true

FredLL-Avaiga (approved):

src/taipy/gui/viselements.json
FabienLelaquais: If you say so. One of these days I will be able to do what you just did.

Issue #774, opened 2023-06-06
Class API


dinhlongviolin1:
**What would that feature address**
This feature would provide a way to declare taipy-gui visual elements similarly to what Gradio and Streamlit is doing (by using classes). With this feature, users that are familiar with those libarries can easily transition to Taipy.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #801, opened 2023-06-20
Class API (#774)

dinhlongviolin1:
Solve Issue #774 
To build pyi files
```bash
cp tools/generate_pyi.py temp.py && python temp.py && rm temp.py
```

```python
from src.taipy.gui import Gui, ClassApi

import src.taipy.gui as tp

a = "Hey There!"

with tp.part() as block1:
    tp.text(value = "Hello world!", raw=True)
    tp.input(value = "{a}")

gui = Gui()
gui.add_page("page1", ClassApi(block1))

gui.run(run_browser=False)
```


src/taipy/gui/renderers/_class_api/element_api.py
```diff
@@ -0,0 +1,77 @@
...
+        return "\n".join([child._render(gui) for child in self._children])
+
+
+class ControlElementApi(ElementApi):
+    def __init__(self, **kwargs):
+        super().__init__(**kwargs)
+
+    def _render(self, gui: "Gui") -> str:
+        el = _ClassApiFactory.create_element(gui, self.ELEMENT_NAME, self._properties)
+        return f"<div>{el[0]}</{el[1]}></div>"
```
FredLL-Avaiga: do we need key and css class  for the div ?

FredLL-Avaiga (approved):
Well done

src/taipy/gui/renderers/_class_api/element_api.py
dinhlongviolin1: I think we do since this will be push into JSXParser. Can you help me on what to add to the div?

FredLL-Avaiga: at least key should be unique
What is used the markdown postproc is & static method of _Builder that you may be able to re-use ?
https://vscode.dev/github/Avaiga/taipy-gui/blob/develop/src/taipy/gui/renderers/_markdown/postproc.py#L33

Issue #804, opened 2023-06-21
BUG-default value for transform types of default property


FredLL-Avaiga:
**Description**
When setting a value to a default property, the UI doesn't reflect the the value on init if the value is not of a simple type (ie needs a holder)

**How to reproduce**
https://github.com/Avaiga/taipy/issues/222

**Expected behavior**
the value is transformed and the UI initialised

PR #806, opened 2023-06-21
Backport: default value of default property (when type needs holder) (#804)

FredLL-Avaiga:
- bump version to 2.3.1


FabienLelaquais (approved):
Yeah yeah. Same ol' same ol'

Issue #796, opened 2023-06-14
Support natively variable sharing between clients


jrobinAV:
**Description**
The purpose is to enhance the variable sharing between clients.

Here is the use case. 
1. I want to display the total number of scenarios. I am using the `state.scenarios` variable for that.
2. I want to select a scenario in a `scenario_selector`, and display the scenario label. I am using the `state.selected_scenario` variable for that.

``` python
import random

import taipy as tp
from config import s_cfg

scenarios = []
selected_scenario = None

def create_random_scenario(state):
    random_int = random.randint(0, 10000)
    scenario = tp.create_scenario(s_cfg, name=f"{random_int}")
    scenario.b.path = f"data/b__{random_int}.pickle"
    scenario.submit()
    state.scenarios = tp.get_scenarios()

def on_init(state):
    state.scenarios = tp.get_scenarios()

page = """
<|create and submit scenario|button|on_action=create_random_scenario|>

nb of scenarios: <|{len(scenarios)}|text|>

<|{selected_scenario}|scenario_selector|>
"""

gui = tp.Gui(page)
core = tp.Core()

if __name__ == "__main__":
    core.run()
    gui.run()

```
My problem is that I want my variable `state.scenarios` to be shared by all the clients (user sessions) while the `state.selected_scenario` variable to be scoped for each client(user session).

The symptom is described in the next screenshot. The number of scenarios is incorrect except in the browser that created the latest scenario. Note that on the scenario_selector, the scenarios are correctly broadcasted.

![image](https://github.com/Avaiga/taipy-gui/assets/88036007/50fca54f-fae9-4fd7-bd83-05655ffbcaf4)

Here is the config used, but you don't care. It s just to reproduce the use case.

``` python
import random

import taipy as tp
from taipy import Scope


def f(df):
    return random.randint(0, 100)

tp.Config.configure_global_app(read_entity_retry=3)
a_cfg = tp.Config.configure_csv_data_node("a", scope=Scope.GLOBAL, default_path="data/a.csv")
b_cfg = tp.Config.configure_data_node("b")
t_cfg = tp.Config.configure_task("t", function=f, input=a_cfg, output=b_cfg)
s_cfg = tp.Config.configure_scenario_from_tasks("s", [t_cfg])
```

PR #818, opened 2023-06-28
broadcast variable / Variable sharing (#796)

dinhlongviolin1:
solve issue #796 

```python
from src.taipy.gui import Gui

a = 30
b = 50
c = 90

Gui.add_shared_variable("a", "c")

def btn_action(state):
    state.broadcast("b", 10)
    # state.b = 10

gui = Gui("""

<|{a}|number|>

<|{b}|number|>

<|{c}|number|>

<|Update b to 10|button|on_action=btn_action|>
""")

gui.run(run_browser=False)
```


FredLL-Avaiga:
Should this be discussed on a global context ?

src/taipy/gui/data/data_scope.py
```diff
@@ -47,14 +47,6 @@ def get_scope(self, client_id: t.Optional[str]) -> SimpleNamespace:
     def get_all_scopes(self) -> t.Dict[str, SimpleNamespace]:
         return self.__scopes
 
-    def broadcast_data(self, name, value):
-        if self.__single_client:
-            return
-        if not self.__client_id:
-            for _, v in self.__scopes.items():
```
FredLL-Avaiga, line 54: can we use values() here ?

src/taipy/gui/gui.py
```diff
@@ -1541,6 +1546,14 @@ def broadcast(self, name: str, value: t.Any):
         """
         self.__send_ws_broadcast(name, value)
 
+    def _broadcast_all_clients(self, name: str, value: t.Any):
+        setattr(g, "is_broadcasting", True)
+        self._update_var(name, value)
```
FredLL-Avaiga, line 1569: Would this also broadcast expression that depends on the var ?

src/taipy/gui/utils/_attributes.py
```diff
@@ -27,11 +27,19 @@ def _getscopeattr_drill(gui: "Gui", name: str) -> t.Any:
 
 
 def _setscopeattr(gui: "Gui", name: str, value: t.Any):
-    setattr(gui._get_data_scope(), name, value)
+    if gui._is_broadcasting():
+        for _, scope in gui._get_all_data_scopes().items():
```
FredLL-Avaiga: items() => values() ?

FabienLelaquais (approved):
Beside the remarks of Fred, I think that's great.
I was wondering, in terms of simplicity of use, if you could provide something like
    Gui.set_broadcast("<varname>")
If then some callback does a `state.<varname> = value` it would be broadcast automatically...
(I don't yet like my naming)

src/taipy/gui/data/data_scope.py
dinhlongviolin1: certainly yes. Didn't optimize yet. Thanks Fred!

src/taipy/gui/gui.py
dinhlongviolin1: yes it will. `_update_var` will also take into account the revaluation and update the `derived_modified` list, which will propogate to all clients through `__send_ws_update_with_dict` method.

src/taipy/gui/utils/_attributes.py
dinhlongviolin1: thanks!

dinhlongviolin1:
> Beside the remarks of Fred, I think that's great.
> I was wondering, in terms of simplicity of use, if you could provide something like
> Gui.set_broadcast("")
> If then some callback does a `state.<varname> = value` it would be broadcast automatically...
> (I don't yet like my naming)

maybe something like `gui.set_global(var_name)`

FabienLelaquais:
> > Beside the remarks of Fred, I think that's great.
> > I was wondering, in terms of simplicity of use, if you could provide something like
> > Gui.set_broadcast("")
> > If then some callback does a `state.<varname> = value` it would be broadcast automatically...
> > (I don't yet like my naming)
> 
> maybe something like `gui.set_global(var_name)`

How about `Gui.set_shared(var_name: str) -> None`?

src/taipy/gui/gui.pyi
```diff
@@ -208,6 +213,10 @@ class Gui:
     def __bind_local_func(self, name: str): ...
```
FredLL-Avaiga, line 213: a *really* private function here ?

src/taipy/gui/gui.pyi
```diff
@@ -208,6 +213,10 @@ class Gui:
...
     def load_config(self, config: Config) -> None: ...
     def broadcast(self, name: str, value: t.Any): ...
+    def _broadcast_all_clients(self, name: str, value: t.Any): ...
+    def _set_broadcast(self, broadcast: bool = ...): ...
+    def _del_broadcast(self) -> None: ...
+    def _is_broadcasting(self) -> bool: ...
```
FredLL-Avaiga, line 219: do we need to make those available in gui.pyi ?

FredLL-Avaiga (approved):

src/taipy/gui/gui.pyi
dinhlongviolin1: it has been for the last 18 months :D

src/taipy/gui/gui.pyi
dinhlongviolin1: a script generates that. we can optionally remove those but I think it is ok to keep it. Pyi should be kept as an alias to the original file

Issue #822, opened 2023-07-03
BUG- changing the table column names does not reset the filters


Dr-Irv:
**Description**
If you have the ability to switch between tables using a selector, and the user applies a filter to one table, when you switch to the second table, the filter remains, and an error message is produced.

**How to reproduce**
A code fragment
```python
import pandas as pd

from taipy.gui import Gui, Markdown

table1 = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
table2 = table1 * 10
table2.columns = ["a", "b"]
selected = "table1"
thetable = table1
lov = ["table1", "table2"]

page = Markdown(
    """
<|{selected}|selector|lov={lov}|dropdown|on_change={update_table}|>

<|{thetable}|table|filter=True|rebuild=True|width=fit-content|>
"""
)


def update_table(state):
    if state.selected == "table1":
        state.thetable = table1
    elif state.selected == "table2":
        state.thetable = table2


if __name__ == "__main__":
    gui = Gui(page=page)
    gui.run(title="Filter Test")
```

A sequence of events:

1. Click the filter button on table1
2. Enter a filter "x equals 2"
3. Use the selector to choose table2

You will get the message:
```text
TaipyGuiWarning: Dataframe filtering: invalid query '`x` == 2' on     a   b
0  10  40
1  20  50
2  30  60:
name 'x' is not defined
```

That's because "table2" has different column names

**Expected behavior**
Ideally, if a filter existed, and the table changes, then the filter should remain ONLY for column names that are present in the new table.  This would allow the same filter to be applied on other tables without having to change the filter, if the column names matched up. 

However, an alternate solution would be to clear the filter when the column names of the table have changed.

**Screenshots**
When available and relevant, screenshots better help show the problem.

**Runtime environment**
Please specify relevant indications.
 - OS: Windows 10 22H2
 - Browser: Chrome

taipy 2.3.1

FlorianJacta:
Thank you Irv for your issue. We will investigate it and keep in touch!

PR #823, opened 2023-07-04
Handle table filters on column change

FredLL-Avaiga:
Addresses #822.

- adding taipy-busy class name to activity spinner


dinhlongviolin1 (approved):
Obviously u're the best

FabienLelaquais (approved):
What else?

PR #829, opened 2023-07-10
[Backport] Adapt table filters to visible columns

FredLL-Avaiga:
Addresses #822.

* filter init add new test on unapplicable filter
* taipy-busy css class name added to activity spinner


FabienLelaquais (approved):
👍

Issue #824, opened 2023-07-06
BUG-Multiline Input Visual Element, can't break line


AlexandreSajus:
**Description**
Using Enter or Shift+Enter while editing text in a multiline input visual element triggers on_action but does not break line

**How to reproduce**
Create a Taipy page with a multiline input visual element

![image](https://github.com/Avaiga/taipy-gui/assets/62465003/70dc0a46-27e8-4899-9d17-cc02606d2f4e)

![image](https://github.com/Avaiga/taipy-gui/assets/62465003/1797ebf9-f1de-47ec-964f-9c8890ba8946)

Press Enter or Shift+Enter while editing the text inside the element

**Expected behavior**
Shift+Enter should break line when editing text in a multiline input element

**Runtime environment**
Please specify relevant indications.
 - OS: Windows 11
 - Browser: Chrome

FredLL-Avaiga:
Does it work if on_action is not defined ?

FredLL-Avaiga:
it would make life easier if you could include a reproduction code example so that one can copy/paste.

AlexandreSajus:
It does work when on_action is not defined. But even if on_action is defined, Shift+Enter should still break line

Sure! Here's a reproduction code:

```python
from taipy.gui import Gui

report = ""


def report_feedback(state):
    return None


page = """
<|{report}|input|on_action=report_feedback|multiline|>
"""

gui = Gui(page)
gui.run(title="Talk To Taipy (alpha)")
```

FredLL-Avaiga:
Does this need to be backported to 2.3.* ?
@FabienLelaquais

FredLL-Avaiga:
backporting to 2.3

PR #830, opened 2023-07-10
ignore shift-alt-ctrl keys with on_action (#824)

FredLL-Avaiga:
```
from taipy.gui import Gui

report = ""


def report_feedback(state):
    return None


page = """
<|{report}|input|on_action=report_feedback|multiline|>
"""

gui = Gui(page)
gui.run(title="Talk To Taipy (alpha)")
```

### DIFF
--- a/gui/src/components/Taipy/Input.tsx
+++ b/gui/src/components/Taipy/Input.tsx
@@ -22,12 +22,14 @@ import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../..
 const AUTHORIZED_KEYS = ["Enter", "Escape", "F1", "F2", "F3", "F4", "F5", "F6", "F7", "F8", "F9", "F10", "F11", "F12"];
 
 const getActionKeys = (keys?: string): string[] => {
-    const ak = keys
-        ? keys
-              .split(";")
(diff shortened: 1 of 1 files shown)


gui/src/components/Taipy/Input.tsx
```diff
@@ -22,12 +22,14 @@ import { useClassNames, useDispatch, useDynamicProperty, useModule } from "../..
...
-              .map((v) => v.trim())
-        : [];
+    const ak = (
+        keys
+            ? keys
+                  .split(";")
+                  .map((v) => v.trim().toLowerCase())
+                  .filter((v) => AUTHORIZED_KEYS.some((k) => k.toLowerCase() === v))
+            : []
+    ).map((v) => AUTHORIZED_KEYS.find((k) => k.toLowerCase() == v) as string);
```
FabienLelaquais, line 32: Isn't this `map` useless the `keys` is empty?

FabienLelaquais (approved):
👌

gui/src/components/Taipy/Input.tsx
FredLL-Avaiga: won't add much if the array is empty :-)
and otherwise the right case is used

Issue #832, opened 2023-07-11
Upgrade for Pandas 2.0 support


FabienLelaquais:
**Description**

To align with Core... and Taipy.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #833, opened 2023-07-11
Upgrade Pandas dependency to version 2.0

FabienLelaquais:
Addresses #832.

Pandas minimal required version is now set to 2.0 (to match with taipy-core requirements)


setup.py
```diff
@@ -33,7 +33,7 @@
     "flask-cors>=3.0.10,<4.0",
     "flask-socketio>=5.3.0,<6.0",
     "markdown>=3.4.1,<4.0",
-    "pandas>=1.5.1,<2.0",
+    "pandas>=2.0.0,<3.0",
```
jrobinAV, line 36: modin specifies `pandas>=2,<2.1` I would stick to the same.

toan-quach (approved):
lgtm, I agree with JR remark also

setup.py
FabienLelaquais: > modin specifies `pandas>=2,<2.1` I would stick to the same.

Hum... I have trouble setting a new kind of restriction due to a package we don't even use.
I do see the point, but I would find it difficult to explain to Taipy GUI (only) users.

If one installs Taipy, the restrictions will apply properly, which is the objective.

Issue #825, opened 2023-07-06
BUG-on_init called too early


FredLL-Avaiga:
**Description**
user function on_init should be called after library initialisations

**How to reproduce**
add an on_init function to your script that uses a library
if a state modification has an impact on a library component behavior, an exception is raised

**Expected behavior**
on_init should be allowed to have an impact on library components

FredLL-Avaiga:
The issue seems more complicated than this.
On first render, the core gui components request inner state variables that are defined by library.on_init.
library.on_init is called during the first ws call (init_router) but the components are generating ws call as soon as they are rendered ie not waiting for init_router return.
llibrary.on_init should therefore be called earlier ie on first render of a page ...

PR #836, opened 2023-07-12
Extension: Library.on_init() must be invoked earlier

FredLL-Avaiga:
Addresses #825.

Make sure library.on_init() is called only once


src/taipy/gui/gui.pyi
```diff
@@ -219,6 +220,7 @@ class Gui:
     def _hold_actions(self, callback: t.Optional[t.Union[str, t.Callable]] = ..., message: t.Optional[str] = ...): ...
     def _resume_actions(self) -> None: ...
     def _navigate(self, to: t.Optional[str] = ..., tab: t.Optional[str] = ..., force: t.Optional[bool] = ...): ...
+    def __init_libs(self) -> None: ...
```
FabienLelaquais, line 222: Do we really want the `_` functions and methods in the pyi file?

FabienLelaquais (approved):
Tricky one this one. Well done!

src/taipy/gui/gui.pyi
FredLL-Avaiga: I don't think so but @dinhlongviolin1 knows better ;-)

dinhlongviolin1: Lets keep it for now, will come up with new process later (probably this week)

dinhlongviolin1 (approved):

Issue #838, opened 2023-07-12
BUG- CSS file not discovered by Taipy Enterprise


FlorianJacta:
**Description**
The CSS file (named main.css) is not discovered by Taipy Enterprise by default but is discovered by Taipy Community. It can have other caveats in Taipy Gui Enterprise.

**How to reproduce**
main.py
```python
from taipy.gui import Gui

data = {"x":[1, 2, 3, 4, 5, 6],
        "y":[12, None, 31, 4, 8, 6],}

md = """
<|{data}|chart|x=x|y=y|width=400%|>
"""

Gui(md).run()
```

main.css
```
.taipy-chart {
      overflow: auto; /* unset*/
}
```

**Expected behavior**
The CSS file should be discovered and used by Taipy Enterprise. Other caveats related to this issue should be fixed.

**Runtime environment**
Taipy Enterprise 2.3.1

PR #840, opened 2023-07-12
Gui when used by the Enterprise edition does not find the main CSS file

FredLL-Avaiga:
Addresses #838.

So that Enterprise Edition can use the CSS file resolution mechanism.


FabienLelaquais (approved):
🤔

dinhlongviolin1 (approved):
LGTM

PR #841, opened 2023-07-12
save the index when a filter is applied

### LINKED ISSUE #835: BUG- When a table is filtered, `on_action()` does not pass the correct row number of the original DataFrame
**Description**
If a user filters a table, and an `on_action()` callback is set on the table, then the `payload["index"]` contains the row number of the filtered table, but there is no way to know what real row that is in the original `DataFrame`.

We are pretty sure that the same error occurs with `on_edit()` as well.

**How to reproduce**
```python
from typing import Any

import pandas as pd

from taipy.gui import Gui, Markdown, State

table1 = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})

indexval = -1
xval = "-1"
yval = "-1"

page = Markdown(
    """

<|{table1}|table|filter=True|rebuild=True|width=fit-content|on_action=on_action|>

<|Indexval|text|> <|{indexval}|text|>

<|xval|text|> <|{xval}|text|>

<|yval|text|> <|{yval}|text|>

"""
)


def on_action(state: State, var_name: str, action: str, payload: dict[str, Any]):
    indexval = payload["index"]
    xval = state.table1.iloc[indexval]["x"]
    state.xval = str(xval)
    yval = state.table1.iloc[indexval]["y"]
    state.yval = str(yval)
    state.indexval = indexval


if __name__ == "__main__":
    gui = Gui(page=page)
    gui.run(title="Filter Action")

```

Sequence of events:
- Filter the table on "x" "equals" "3"
- That will leave one row.  Choose that row
- The action updates the index values and the corresponding row of the DataFrame, which is not the filtered value

**Expected behavior**
The row number should be of the filtered DataFrame, or there should be some way to know what rows remain after filtering.

**Runtime environment**
Please specify relevant indications.
 - OS:  Windows 10 2H22
 - Browser: Chrome  114.0.5735.134
 - taipy 2.3.1

Dr-Irv:
Fixes #835 

Note - I had to change typing issues in 2 of the files to get the pre-commit to work

### DIFF
--- a/src/taipy/gui/data/pandas_data_accessor.py
+++ b/src/taipy/gui/data/pandas_data_accessor.py
@@ -234,6 +234,10 @@ class _PandasDataAccessor(_DataAccessor):
                 value = value.query(query)
             except Exception as e:
                 _warn(f"Dataframe filtering: invalid query '{query}' on {value.head()}:\n{e}")
+            if _PandasDataAccessor.__INDEX_COL not in value.columns:
+                value[_PandasDataAccessor.__INDEX_COL] = value.index
+            if _PandasDataAccessor.__INDEX_COL not in columns:
+                columns.append(_PandasDataAccessor.__INDEX_COL)
 
         if paged:
             aggregates = payload.get("aggregates")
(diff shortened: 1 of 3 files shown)


FredLL-Avaiga:
I came up with the same fix :-D

FabienLelaquais:
Closed as a duplicate of #871

PR #850, opened 2023-07-18
Add multiple page modules from folder

dinhlongviolin1:


src/taipy/gui/gui.py
```diff
@@ -1236,17 +1239,29 @@ def _taipy_on_cancel_block_ui(guiApp, id: t.Optional[str], payload: t.Any):
...
+            elif re_match := Gui.__RE_PY.match(file_name):
+                module_name = re_match.group(1)
+                module_path = os.path.join(folder_name, module_name).replace(os.path.sep, ".")
+                try:
+                    module = importlib.import_module(module_path)
+                    page_instance = _get_page_from_module(module)
+                    if page_instance is not None:
+                        self.add_page(name=f"{folder_name}/{module_name}", page=page_instance)
+                except Exception as e:
+                    warnings.warn(f"Error while importing module '{module_path}': {e}")
```
FredLL-Avaiga: Should we use _warn from _warnings.py ?

src/taipy/gui/gui.pyi
```diff
@@ -37,6 +37,7 @@ class Gui:
     __DO_NOT_UPDATE_VALUE: Incomplete
     __RE_HTML: Incomplete
     __RE_MD: Incomplete
+    __RE_PY: Incomplete
```
FredLL-Avaiga, line 40: Dr Irv had propsed a change about this Incomplete  (change the import to from typing import Any as Incomplete)
Do you have any take on this ?
I guess this is generated and cannot be changed ?

FredLL-Avaiga (approved):
🥇

FredLL-Avaiga:
Should merge @trgiangdo to fix the checks

src/taipy/gui/gui.pyi
dinhlongviolin1: we cannot change this as it is generated by stubgen. we can override this manually but I dont see the point of us doing it

src/taipy/gui/gui.py
dinhlongviolin1: thanks Fred!

src/taipy/gui/gui.py
```diff
@@ -206,6 +208,7 @@ class Gui:
 
     __RE_HTML = re.compile(r"(.*?)\.html")
```
FabienLelaquais: In those three regexps, I suspect you should have an additional '$' at the end of the regexp.
Otherwise these will match "myfile.md.text", "file.pyi", and others...

FabienLelaquais (approved):
Love it.

src/taipy/gui/gui.py
dinhlongviolin1: thanks!

PR #859, opened 2023-07-21
update pyi + release publish process (#860)

dinhlongviolin1:
Solve Issue #860


FredLL-Avaiga (approved):
So the pyi is not committed anymore, it's only generated and included in the release?

dinhlongviolin1:
> So the pyi is not committed anymore, it's only generated and included in the release?

Yes you are right!

dinhlongviolin1:
> So the pyi is not committed anymore, it's only generated and included in the release?

And private methods and attributes will not be included in the pyi files

FabienLelaquais (commented):
👌

PR #868, opened 2023-07-24
Update release process

dinhlongviolin1:
- Keep the dev and normal version release process separated (since both have different input and release workflow requirements)
- Inputs for dev release (the version of the dependencies) will not be required per request from JR
- Merge publish process for both dev and normal version


FredLL-Avaiga (approved):

.github/workflows/release-dev.yml
```diff
@@ -1,15 +1,14 @@
...
     inputs:
       taipy-config-version:
-        description: "The taipy-config version to use (ex: 2.3.0.dev0)"
-        required: true
+        description: "The taipy-config version to use (ex: 1.0.0.dev0)"
 
 jobs:
-  publish:
-    timeout-minutes: 20
+  test-release-package:
```
jrobinAV: I don't understand the prefix test. Why not:
```suggestion
  release-dev-package:
```

.github/workflows/release-dev.yml
```diff
@@ -52,10 +51,6 @@ jobs:
           """ > /tmp/check1.py
           python /tmp/check1.py
 
-      - name: Check dependencies are available
-        run: |
-          curl https://pypi.org/simple/taipy-config/ | grep -o ">taipy-config-${{ github.event.inputs.taipy-config-version }}\.tar\.gz<"
-
       - name: Ensure package version is properly set
```
jrobinAV:
```suggestion
      - name: Extract package version
```

.github/workflows/publish.yml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "The Pypi package version to create (ex: 1.0.0)"
+        description: "The Pypi package version to create (ex: 1.0.0, 1.0.0.dev0)"
```
jrobinAV:
```suggestion
        description: "The tag of the package to publish on Pypi (ex: 1.0.0, 1.0.0.dev0)"
```

.github/workflows/release.yml
```diff
@@ -8,8 +8,8 @@ on:
         required: true
 
 jobs:
-  test-package:
-    timeout-minutes: 20
+  test-release-package:
```
jrobinAV: I don't understand this name. Why the prefix test?

.github/workflows/publish.yml
```diff
@@ -82,7 +77,7 @@ jobs:
 
   test-published-package:
     needs: [publish-to-pypi]
-    timeout-minutes: 30
+    timeout-minutes: 40
```
jrobinAV, line 80: 40 minutes for pip install taipy-gui? Ouch 😢

jrobinAV (commented):
On release-dev:
1 - I would rename "Ensure package version is properly set" to "Extract package version" for consistency with other steps.

2 - I would remove the step "Validate branch name (has 'dev' prefix)", so we can build a dev release from any branch. 
For instance, I would like to package a dev release on a new feature branch like `feature/#42-add-new-feature`, so I can test it before merging into develop. Does it make sense to you?

3 - I am not sure to understand the impact of the step "Increase dev version" when the branch is not `develop` but `dev/2.5` or eventually `feature/#42-add-new-feature` if we remove the "Validate branch name" step (cf second point).

FabienLelaquais (approved):
👌

jrobinAV (approved):

Issue #866, opened 2023-07-24
BUG- A pandas DataFrame with a non-contiguous index has issues when editing


Dr-Irv:
**Description**
If you have a pandas DataFrame where there are holes in the index, then the value of `payload["index"]` in the `on_edit()` function is correct if the DataFrame is sorted, but incorrect if not.

**How to reproduce**
```python
import traceback
from taipy import Gui
from taipy.gui.state import State
import pandas as pd

df = pd.DataFrame(
    {"lets": list("ABCDEFGHIJ"), "num": [i + 1 for i in range(10)]},
    index=[(i + 1) * 10 + i for i in range(10)],
)


def myedit(state: State, var_name: str, action: str, payload: dict):
    try:
        print("incoming df")
        print(state.df)
        print("index ", payload["index"])
        print("column ", payload["col"])
        print("value ", payload["value"])
        print("row from index")
        print(state.df.loc[payload["index"]])
        state.df.at[payload["index"], payload["col"]] = payload["value"]
        print("changed ")
        print(state.df)
        state.df = state.df
    except Exception:
        print("EXCEPTION ")
        traceback.print_exc()


page = "<|{df}|table|id=df|on_edit=myedit|editable=False|editable[num]|width=300px|>"

if __name__ == "__main__":
    Gui(page=page).run()
```

Click on the 3rd row (labeled "C") and edit the value "3" to be "35".  You will get an exception that `payload["index"]` is not found, because the value is `2`, but the index is `[11, 22, 33, ..., 99, 110]`

Then sort on the  `let` column.  Try the edit again, and it works.

**Expected behavior**
The `payload["index"]` value should be the true index value, whether sorted or not.

taipy 2.3.1

May be related to #841

PR #871, opened 2023-07-26
Copy DataFrame only once

FredLL-Avaiga:
Addresses #866.

dataframe was notcopied when adding index
but it should only be copied once !


dinhlongviolin1 (approved):
Looks good to me!

FabienLelaquais (approved):
😶

Issue #867, opened 2023-07-24
BUG- In a table, the date format is not respected when editing


Dr-Irv:
**Description**
If you have date/times in your table, and you specify a date format, taipy presents a full date/time to edit, even if you only want the date.

**How to reproduce**
A code fragment
```python
import traceback
from taipy import Gui
from taipy.gui.state import State
import pandas as pd

dr = pd.date_range(start="3/1/2023", periods=10, freq="D")
df = pd.DataFrame(
    {"dates": list(dr)}
)


def myedit(state: State, var_name: str, action: str, payload: dict):
    state.df.at[payload["index"], payload["col"]] = payload["value"]
    state.df = state.df


page = "<|{df}|table|id=df|on_edit=myedit|editable=True|width=250px|date_format=MM/dd/yy|>"

if __name__ == "__main__":
    Gui(page=page).run()
```

Edit any date, which is in the format `MM/dd/yy`, and you will get a string to edit that looks like this: `2023-03-01T05:00:00.000000Z`

**Expected behavior**
If you specify a date format, and you make the value editable, then it should respect that date format!

**Screenshots**

![image](https://github.com/Avaiga/taipy-gui/assets/15113894/3b28f966-921c-48ca-8488-7ea1bbef913f)

**Runtime environment**
Windows 10, Chrome (latest), taipy 2.3.1

Related to #797, but fixing this editing now would be very helpful!

Issue #797, opened 2023-06-15
Allow date picker for dates in tables


Dr-Irv:
**What would that feature address**
We are displaying a table that has editable columns that include dates.  Right now, if you leave it as a pandas datetime type, it prompts to enter a very long string.  In our case, we just want to let the user pick a date.

***Description of the ideal solution***
We'd like it that if someone clicks on a date field in a table, it pops up a date picker.  

***Caveats***
Just makes a better user interfacE!

***Other options***
Not sure.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

FredLL-Avaiga:
date values are sent to the callback as an UTC ISO string (with additionnal tz) that needs to be parsed to a date
```
from taipy.gui import Gui, State
import pandas as pd
from dateutil import parser
from zoneinfo import ZoneInfo

dr = pd.date_range(start="3/1/2023", periods=10, freq="D")
df = pd.DataFrame(
    {"dates": list(dr)}
)


def myedit(state: State, var_name: str, action: str, payload: dict):
    val = parser.parse(payload["value"]).astimezone(ZoneInfo(payload["tz"])).replace(tzinfo=None) if payload["col"] == "dates" else payload["value"]
    state.df.at[payload["index"], payload["col"]] = val
    state.refresh("df")


page = "<|{df}|table|id=df|on_edit=myedit|date_format=dd/MM/yy|>"

if __name__ == "__main__":
    Gui(page=page).run()

```

PR #872, opened 2023-07-26
Use date picker when editing date column (#797 #867)

FredLL-Avaiga:
Addresses #797 and #867.

on_edit callback should handle the date column depending on the type of datetime used in the dataframe
```
dr = pd.date_range(start="3/1/2023", periods=10, freq="D")
df = pd.DataFrame(
    {"dates": list(dr)}
)

    val = parser.parse(payload["value"]).astimezone(ZoneInfo(payload["tz"])).replace(tzinfo=None) if payload["col"] == "dates" else payload["value"]
    state.df.at[payload["index"], payload["col"]] = val
    state.refresh("df")

page = "<|{df}|table|id=df|on_edit=myedit|width=250px|date_format=dd/MM/yy|>"

```

### DIFF
--- a/gui/src/utils/index.ts
+++ b/gui/src/utils/index.ts
@@ -58,6 +58,26 @@ interface StyleKit {
     inputButtonHeight: string;
 }
 
+// return date with right time and tz
+export const getTimeZonedDate = (d: Date, tz: string, withTime: boolean) => {
+    const newDate = d;
+    // dispatch new date which offset by the timeZone differences between client and server
+    const hours = getClientServerTimeZoneOffset(tz) / 60;
+    const minutes = getClientServerTimeZoneOffset(tz) % 60;
+    newDate.setSeconds(0);
+    newDate.setMilliseconds(0);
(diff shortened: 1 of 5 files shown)


gui/src/utils/index.ts
```diff
@@ -67,7 +87,7 @@ export const getDateTime = (value: string | null | undefined, tz?: string): Date
         return null;
     }
     try {
-        return (tz && tz !== "Etc/Unknown") ? utcToZonedTime(value, tz) : new Date(value);
+        return tz && tz !== "Etc/Unknown" ? utcToZonedTime(value, tz) : new Date(value);
```
FabienLelaquais, line 90: I do enjoy the parenthesis that in my opinion make the code more readable.
No big deal...

FabienLelaquais (approved):
Excellent!

gui/src/utils/index.ts
FredLL-Avaiga: but the formatter removes them!

PR #874, opened 2023-07-27
backport: date picker when editing date column (#797 #867)

FredLL-Avaiga:
```
import pandas as pd
from dateutil import parser
from zoneinfo import ZoneInfo

dr = pd.date_range(start="3/1/2023", periods=10, freq="D")
df = pd.DataFrame(
    {"dates": list(dr)}
)


    val = parser.parse(payload["value"]).astimezone(ZoneInfo(payload["tz"])).replace(tzinfo=None) if payload["col"] == "dates" else payload["value"]
    state.df.at[payload["index"], payload["col"]] = val
    state.refresh("df")


page = "<|{df}|table|id=df|on_edit=myedit|date_format=dd/MM/yy|>"

if __name__ == "__main__":
    Gui(page=page).run()

```


dinhlongviolin1 (approved):

FabienLelaquais (approved):
👍

PR #877, opened 2023-07-28
update package.lock.json

FredLL-Avaiga:


FabienLelaquais (approved):
I would have done exactly that.

PR #884, opened 2023-08-09
export table and chart

FredLL-Avaiga:


dinhlongviolin1 (approved):
Linting doing heavy lifting for the PR

PR #889, opened 2023-08-16
Extension: allow property replacement

FredLL-Avaiga:
allows to define a property as part of an expression
ie <tp:prop:...property name...> will be replaced by the property value or expression or None if it is not defined

### DIFF
--- a/src/taipy/gui/extension/library.py
+++ b/src/taipy/gui/extension/library.py
@@ -159,9 +159,9 @@ class Element:
                 if val:
                     # handling property replacement in inner properties <tp:prop:...>
                     while m := Element.__RE_PROP_VAR.search(val):
-                        if (var:= attributes.get(m.group(1))):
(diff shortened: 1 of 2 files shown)


dinhlongviolin1 (approved):

src/taipy/gui/extension/library.py
```diff
@@ -87,6 +88,8 @@ class Element:
     what the default property name is.
     """
 
+    __RE_PROP_VAR = re.compile(r"<tp:prop:(\w+)>")
```
FabienLelaquais, line 91: Would this work if the var name is a property in a dict, if this makes sense?
like `my_dict.my_entry`, as '.' is not captured.

Reading the code... it seems you're locating an element property name, that cannot contain dots, so we're good.

FabienLelaquais (approved):
👌

PR #901, opened 2023-08-29
fix(stylekit): normalize the height of the taipy selector dropdown between multiple and single options

ghost:
This PR addresses Florian Jacta's direct demand concerning the issue below where two dropdown selector with different properties had different heights : 
image


FabienLelaquais (approved):
Certainly fine

FredLL-Avaiga (approved):

PR #907, opened 2023-09-04
Feature/taipy#260 - Update _GuiCLI arguments to be reused in taipy run command

trgiangdo:
https://github.com/Avaiga/taipy/issues/260


FabienLelaquais (approved):
Don't really know if this works, but a smart way to deal with the existing code. Well done!

jrobinAV (approved):

Issue #886, opened 2023-08-09
BUG- Missing values in pandas tables are not handled in a consistent way


Dr-Irv:
**Description**
Thanks for the date picker, but we have a new requirement.  We need to be able to support having "no date" (i.e., `pd.NaT`) as an editable value, i.e., the user can pick a date that was unspecified and then enter it, or take an existing date and make it null.

**How to reproduce**
```python
from taipy import Gui
from taipy.gui.state import State
import numpy as np
import pandas as pd

dr = pd.Series(pd.date_range(start="3/1/2023", periods=10, freq="D").tz_localize("UTC"))
# dr.iloc[2] = pd.NaT  DOES NOT WORK
flt = pd.Series(range(10), dtype=float)
flt.iloc[5] = np.nan
iser = pd.Series(range(10), dtype=pd.Int64Dtype())
iser.iloc[7] = pd.NA
df = pd.DataFrame({"dates": dr, "flt": flt, "iser": iser})
print(df.dtypes)


def myedit(state: State, var_name: str, action: str, payload: dict):
    state.df.at[payload["index"], payload["col"]] = payload["value"]
    state.df = state.df
    print("new df ", df)


page = (
    "<|{df}|table|id=df|on_edit=myedit|editable=True|width=400px|date_format=MM/dd/yy|>"
)

if __name__ == "__main__":
    Gui(page=page).run()

```

The DataFrame has 3 columns.  The first is the datetime column.  There is not a way to null it out.  Also, if you put back the line that says ` dr.iloc[2] = pd.NaT `, then taipy has an error when it tries to display the column, because it doesn't know how to make a blank date.

For the floating column, everything works great.  You can edit a missing value, and you can also erase a value and it will become missing.  This is the behavior I would like with dates.

Finally, for the column with the pandas `Int64Dtype`, missing values are represented by `pd.NA`.  This column is not editable at all because taipy treats the column as a string, rather than as an integer (which is what it should do).

**Expected behavior**
The newer pandas types should be supported with `pd.NA` and `pd.NaT`.  You should be able to clear out a date value, or take a missing date value and be able to enter it.

**Runtime environment**
Please specify relevant indications.
Chrome on Windows.  taipy 2.4.0

PR #911, opened 2023-09-11
Missing values in pandas

FredLL-Avaiga:
Address #886.

```

import numpy as np
import pandas as pd

dr = pd.Series(pd.date_range(start="3/1/2023", periods=10, freq="D").tz_localize("UTC"))
dr.iloc[2] = pd.NaT  # DOES NOT WORK
flt = pd.Series(range(10), dtype=float)
flt.iloc[5] = np.nan
iser = pd.Series(range(10), dtype=pd.Int64Dtype())
iser.iloc[7] = pd.NA
df = pd.DataFrame({"dates": dr, "flt": flt, "iser": iser})
print(df.dtypes)


    state.df.at[payload["index"], payload["col"]] = payload["value"]
    state.df = state.df
    print("new df ", df)


page = "<|{df}|table|id=df|on_edit=myedit|editable=True|width=400px|date_format=MM/dd/yy|>"

if __name__ == "__main__":
    Gui(page=page).run()

```


FabienLelaquais (approved):
Well done!

Issue #592, opened 2023-01-30
Create `navbar` like functionality to change parts/partials


Dr-Irv:
It would be nice if `navbar` like functionality could be used to "navigate" to a part with a partial. I'd like to have a navbar at the top of the page that navigates to a page, but then within that page, have a navbar that changes a partial within the page.

FredLL-Avaiga:
Could you use a [toggle ](https://docs.taipy.io/en/latest/manuals/gui/viselements/toggle/)?

Dr-Irv:
> Could you use a [toggle ](https://docs.taipy.io/en/latest/manuals/gui/viselements/toggle/)?

Maybe?  The docs don't say that there is an `on_action()` callable that would get called when the value is changed.  Also, I'd probably have to do a bit of work to make it look as good as the navbar does!

FredLL-Avaiga:
you can use a specific on_change that will be called on toggle
At least you'll have selection indication.

Dr-Irv:
> you can use a specific on_change that will be called on toggle At least you'll have selection indication.

Doesn't the `on_change` then need to be documented on the `toggle` page?

FredLL-Avaiga:
It should indeed!

FredLL-Avaiga:
this should work
```
from taipy.gui import Gui, State, navigate


lov = [("page1", "Page 1"), ("page2", "Page 2"), ("page3", "Page 3")]
value = lov[0]


def on_change(state: State, var_name: str, value: tuple):
    if var_name == "value":
        navigate(state, value[0])


md = """
# Toggle to Navigate

<|{value}|toggle|lov={lov}|>
"""

Gui(md).run()
```

Dr-Irv:
So the `toggle` works, but can you make it style like the navbar?

Also, the docs should be updated to show that `on_change()` applies.

FredLL-Avaiga:
you can use css to give toggle a navbar look:

`<|{value}|toggle|lov={lov}|class_name=navbar|on_change=my_on_change|>`

```
.navbar .MuiButtonBase-root {
    border: 0;
    color: rgba(255, 255, 255, 0.7);
}
.navbar .MuiButtonBase-root.Mui-selected {
    border: 0;
    border-bottom: 1px solid;
    border-color: var(--color-primary);
    color: var(--color-primary);
}

```

FredLL-Avaiga:
call toggle-navbar will be created in the stylekit

PR #915, opened 2023-09-14
Make toggle elements appear as navbars

FredLL-Avaiga:
Addresses #592.

Make toggle look like navbar via class toggle-navbar (in stylekit)

### DIFF
--- a/gui/public/stylekit/controls/toggle.css
+++ b/gui/public/stylekit/controls/toggle.css
@@ -45,3 +45,14 @@
 .taipy-toggle.nolabel .MuiTypography-root {
     display: none;
 }
+
+.toggle-navbar .MuiButtonBase-root {
+    border: 0;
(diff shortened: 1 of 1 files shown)


dinhlongviolin1 (approved):

gui/public/stylekit/controls/toggle.css
```diff
@@ -45,3 +45,17 @@
 .taipy-toggle.nolabel .MuiTypography-root {
     display: none;
 }
+
+.toggle-navbar .MuiButtonBase-root {
```
FabienLelaquais: Should't we protect that rule so it applies only in the taipy-toggle CSS class context?

FabienLelaquais (changes requested):
Please

FabienLelaquais (approved):
👍

PR #918, opened 2023-09-18
Deselect long option in selector

FredLL-Avaiga:
Addresses #917.

- update dependencies
- fix builder bug on dict instanvce check


FabienLelaquais (approved):
👌

Issue #916, opened 2023-09-16
BUG- file_download control doesn't allow dynamic files to be downloaded


Dr-Irv:
**Description**
The `file_download` control is broken.  What we want to do is have someone click the button, then the action is called to create the file, and then that file is downloaded, with the specified name.

What happens is that the file that is downloaded is the first one that was there, and then the action function is called.

Also, the `name` field seems to be ignored.

Ideally, what we would like is to create a temporary file in the file system and then after it is downloaded, the file is deleted. But we could use a `io.BytesIO` object as well, and just pass that, but if we do that the file is called `TaipyContent.bin`

**How to reproduce**

```python
import datetime as dt
from taipy import Gui
import pandas as pd

df = pd.DataFrame({"x": [1, 2, 3], "y": [4, 5, 6]})
filename = "foobar.xlsx"
df.to_excel(filename)


def action(state):
    tfilename = "tmp-" + dt.datetime.now().strftime("%H-%M-%S") + ".xlsx"
    df.to_excel(tfilename)
    state.filename = tfilename
    return


page = "<|{filename}|file_download|name=junk.xlsx|on_action=action|>"

if __name__ == "__main__":
    Gui(page=page).run()
```

**Expected behavior**
What should happen here is that the file that is created dynamically (which has a name based on the hour/minute/second the action function is called) is the one that is downloaded. But what happens is that the file that is downloaded is the *previous* filename, not the newly created one.

In addition, the name of `junk.xlsx` is not respected at all.

**Runtime environment**
Please specify relevant indications.
 - OS:  Windows 10 and Linux
 - Browser: Chrome

taipy 2.4.0

FlorianJacta:
Thank you for the issue, Irv! Indeed, it will be ideal to have the _on_action_ callback called before downloading the file. Having it called after is also useful to delete the file for example. We will look into that!

FredLL-Avaiga:
I suppose the easiest way to do this would be to have a standard button with an action  that would invoke [download(state, ...)](https://docs.taipy.io/en/latest/manuals/reference/pkg_taipy.gui/#functions) on the backend
after creating the right file

Dr-Irv:
> I suppose the easiest way to do this would be to have a standard button with an action that would invoke [download(state, ...)](https://docs.taipy.io/en/latest/manuals/reference/pkg_taipy.gui/#functions) on the backend after creating the right file

[...]

PR #920, opened 2023-09-19
file_download: on_action must be called after download

FredLL-Avaiga:
Addresses #916.

on_action is now called when the file has been downloaded
for gui_actions.download and file_download control


dinhlongviolin1 (approved):
That is really cool

FabienLelaquais (approved):
👌

Issue #932, opened 2023-09-26
BUG-Charts briefly resets to empty when refreshed


gbritoda:
**Description**
Whenever a chart is refreshed, it resets quickly to an empty chart before updating. While this is not very noticeable in faster systems, if there's an slower connection to the GUI or the system is running slow due to handling a lot of data or something, it gets more noticeable.
It also gets very noticeable if you have fast updating data, like a live graph tracking something, which is my example below:

**How to reproduce**
In this code, we should have a continuous sine wave on the graph updating every 0.1 second when the button is pressed.
But since the graphs blinks every update (the need to refresh is due to Avaiga/taipy#466), it's very hard to read the labels

```python
from taipy.gui import Gui
from time import sleep
from math import sin

data = {
        "x":[],
        "y":[]
    }

page = """
<|button|label=start|on_action=start_button_pressed|>
<|{data}|chart|x=x|y=y|rebuild=True|propagate|>
"""

def start_button_pressed(state):
    global data
    x = 0
    while True:
        data['x'].append(x)
        data['y'].append(sin(x))
        state.refresh("data")
        x+=1
        sleep(0.1)

gui = Gui(page=page)

gui.run(
    host="0.0.0.0",
    port=1234,
    debug=True,
    dark_mode=False,
    use_reloader=True
)

```
**Expected behavior**
A continuously flowing graph of a sine wave

**Runtime environment**
Please specify relevant indications.
 - OS: Ubuntu 22.04.2 LTS on WSL
 - Browser: Firefox and Chrome
 - Taipy 2.4.0

FlorianJacta:
Could you help me understand the issue at hand? Is the behavior you describe only visible with many points, or is it only due to a change of data?

A part of your issue could be that Taipy will change its x-axis and y-axis depending on the data. You can make the chart axes static. However, I believe you are talking about labels of data points.

Here is an example of code with static axes. Would it help?

```python
from taipy.gui import Gui
from time import sleep
from math import sin

data = {
        "x":[],
        "y":[]
    }

# my layout to have static axis
layout = {
        "xaxis": {
            "range": [0, 100],  
        },
        "yaxis": {
            "range": [-1, 1], 
        }
    }

page = """
<|button|label=start|on_action=start_button_pressed|>
<|{data}|chart|x=x|y=y|layout={layout}|>
"""

def start_button_pressed(state):
    x = 0
    while x < 100:
        data['x'].append(x)
        data['y'].append(sin(x))

[...]

Issue #864, opened 2023-07-21
BUG-Blank chart after too much changes


FlorianJacta:
**Description**
If a chart's data changes a lot, the corresponding chart will go blank and will not return to its normal behavior without refreshing the page.

**How to reproduce**

```python
from taipy.gui import Gui
from math import cos, exp

value = 10

page = """
Markdown
# Taipy *Demo*

Value: <|{value}|>

<|{value}|slider|>

<|{compute_data(value)}|chart|>
"""


def compute_data(decay:int)->list:
    return [cos(i/6) * exp(-i*decay/600) for i in range(10_000)]


Gui(page).run()
```

Change the slider a lot of times, play with it. The chart should at one point be blank and it is not possible to see the data even by changing the slider again. The user has to refresh the page.

**Expected behavior**
When the chart goes blank, Taipy should try to make it work

**Screenshots**

![sinus](https://github.com/Avaiga/taipy-gui/assets/98709993/1d543d3a-cf55-4c18-961c-5a24de9de206)

**Runtime environment**
Taipy GUI: 2.3.1

FabienLelaquais:
I've been trying to reproduce this problem with Taipy GUI 3.0, watching closely the interaction between the client, that triggers the action of recomputing the data sent to the Chart control, and the server that does the update request.
No errors were detected, and after a while (the time it takes to process all the consecutive change requests from the slider), the chart updates properly. This can, however, take time: there are lots of computations done on the server to deliver the new data.

This delay is obviously caused by the fact that the callback takes a while, preventing immediate feedback on the page. And this is exactly why Taipy recommends that you don't spend too much time on the server responding to front-end events.

There's nothing Taipy GUI can do on its own to fix that delay.
A real problem would occur should the page never update, which I cannot reproduce.

I'm willing to close this issue at this point, after your acknowledgment.

FlorianJacta:
After too many change requests, my chart will stay like this forever. I need to refresh the page to see it the correct chart appears.

![image](https://github.com/Avaiga/taipy-gui/assets/98709993/ad683779-2a2c-4512-b6e9-744576215ba5)

FabienLelaquais:
I understand.
But I cannot reproduce that behavior.
Are you using Taipy 3.0.0.dev*, and would you consider a screen share of that sometime?

FlorianJacta:
I am using today version of Taipy:
taipy-gui              3.0.0.dev3

FlorianJacta:
[...]

PR #935, opened 2023-09-27
Avoid chart flickering when updating data

FabienLelaquais:
Addresses #932 and #864,

Please congrats to Fred. I'm only the scribe at this point.

### DIFF
--- a/gui/src/components/Taipy/Chart.tsx
+++ b/gui/src/components/Taipy/Chart.tsx
@@ -203,6 +203,7 @@ const Chart = (props: ChartProp) => {
     const [selected, setSelected] = useState<number[][]>([]);
     const plotRef = useRef<HTMLDivElement>(null);
     const [dataKey, setDataKey] = useState("__default__");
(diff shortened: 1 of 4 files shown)


gui/src/components/Taipy/Chart.tsx
```diff
@@ -336,8 +337,11 @@ const Chart = (props: ChartProp) => {
     const skelStyle = useMemo(() => ({ ...style, minHeight: "7em" }), [style]);
 
     const dataPl = useMemo(() => {
-        const datum = data && data[dataKey];
-        return datum
+        if (data === null) {
```
FredLL-Avaiga: this gives me an error if I don't set the test as
 data === null && lastDataPl.current

FredLL-Avaiga (approved):

gui/src/components/Taipy/Chart.tsx
FabienLelaquais: Whao

FredLL-Avaiga (approved):

Issue #923, opened 2023-09-21
BUG-Date selector gives non-midnight UTC timestamp after a DST change, even with time_zone="Etc/UTC"


arcanaxion:
**Description**

Background
From my observation, a date selector converts the date selected from the user's time zone to utc. I.e. if I am in UTC+08 and select 2023-09-21, `selected_datetime` would be `dt.datetime(2023, 9, 21, tzinfo=localtz).astimezone(tzutc())`, i.e. `2023-09-20T16:00:00Z`.

By setting the time_zone to utc (`Gui(md).run(time_zone="Etc/UTC")`), I would get `2023-09-21T00:00:00Z`, which is expected. This is convenient for me when using the date selector to get a dt.date (not dt.datetime).

Issue
The odd thing is that if **my machine** is in a timezone which observes DST (e.g. CET/CEST), if I select a date after a DST change, `selected_datetime` will be a non-midnight UTC timestamp -- even though my time_zone is set to "Etc/UTC".

E.g. if my machine is in CEST and the date is 2023-09-21, if I select 2023-09-21 with the date selector, I get `2023-09-21T00:00:00Z`, as expected. But if I select 2023-10-30 (CEST -> CET on 2023-10-29), I get `2023-10-30T01:00:00Z`

**How to reproduce**

```python
import datetime as dt
from taipy.gui import Gui
from dateutil.tz import tzutc

selected_datetime = dt.datetime(2023, 10, 28, tzinfo=tzutc())

md = """
# Timezone issue

<|{selected_datetime}|date|>

<|{repr(selected_datetime)}|>
"""


if __name__ == "__main__":
    gui = Gui(md)
    gui.run(time_zone="Etc/UTC")
```

With the machine timezone set to Amsterdam, try picking a date after a DST change. E.g. since it is currently CEST, try picking 2023-10-30 since that would be CET.

**Expected behavior**
When `time_zone` is set to "Etc/UTC", I expected that the timestamp returned will always be at midnight UTC at any date of any year, regardless of my machine's timezone. This is not the case for a machine with a timezone that observes DST.

**Screenshots**
When available and relevant, screenshots better help show the problem.

**Runtime environment**
Please specify relevant indications.
 - OS: WSL
 - Browser: Chrome
 - Windows timezone set to UTC+01 (Amsterdam). This timezone is observed in WSL

Issue #895, opened 2023-08-23
Option for the "date" control to return a datetime.date object


arcanaxion:
**Description**
Currently, the date control returns a `datetime.datetime` object -- even if the with_time property is False (the default). If I just want the date component, doing `selected_date.date()` may not be sufficient as some added timezone conversion may need to be done.

I would like an option for the date control to return a naive `datetime.date` object.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #936, opened 2023-09-28
Fix timezone (#895, #923)

dinhlongviolin1:
Solve bug #895, #923

A few test run is included

no time date selector

```python
import datetime as dt
from src.taipy.gui import Gui
from dateutil import parser

selected_date = dt.date.today()

md = """
# Timezone issue

<|{selected_date}|date|not with_time|>

<|{selected_date}|>

<|{repr(selected_date)}|>
"""


if __name__ == "__main__":
    gui = Gui(md)
    gui.run(time_zone="Etc/UTC")
```

table with no time
```python
import pandas as pd
from datetime import datetime
import src.taipy.gui.builder as tgb
from src.taipy.gui import Gui

start_date = datetime(2023, 1, 1)
end_date = datetime(2023, 2, 1)
date_range = pd.date_range(start=start_date, end=end_date)

df = pd.DataFrame({'Date': date_range})
df["Index"] = df.index

print(df)

def on_edit(state, var_name, action, payload):
    print(var_name, action, payload)

# with tgb.Page() as page:
#     tgb.table(data="{df}", editable=True, filter=True, on_edit=on_edit)

page = """
<|{df}|table|on_edit=on_edit|editable=True|filter=True|>
"""

Gui(page=page).run()
```

table with time
```python
import pandas as pd
from datetime import datetime
import src.taipy.gui.builder as tgb
from src.taipy.gui import Gui

start_datetime = datetime(2023, 1, 1, 0, 0, 0)
end_datetime = datetime(2023, 1, 3, 23, 59, 59)
datetime_range = pd.date_range(start=start_datetime, end=end_datetime, freq='H')

df = pd.DataFrame({'Date': datetime_range})
df["Index"] = df.index

print(df)

def on_edit(state, var_name, action, payload):
    print(var_name, action, payload)

# with tgb.Page() as page:
#     tgb.table(data="{df}", editable=True, filter=True, on_edit=on_edit)

page = """
<|{df}|table|on_edit=on_edit|editable=True|filter=True|date_format=yyyy-MM-dd HH:mm:ss zzz|>
"""

Gui(page=page).run()
```

### DIFF
--- a/gui/src/components/Taipy/DateSelector.spec.tsx
+++ b/gui/src/components/Taipy/DateSelector.spec.tsx
@@ -143,7 +143,7 @@ describe("DateSelector Component", () => {
             await userEvent.type(input, "{ArrowLeft}{ArrowLeft}{ArrowLeft}01012001", { delay: 1 });
             expect(dispatch).toHaveBeenLastCalledWith({
                 name: "",
-                payload: { value: "2001-01-01T00:00:00.000Z" },
+                payload: { value: "Mon Jan 01 2001" },
                 propagate: true,
                 type: "SEND_UPDATE_ACTION",
             });
--- a/gui/src/utils/hooks.ts
+++ b/gui/src/utils/hooks.ts
@@ -105,7 +105,7 @@ export const useFormatConfig = (): FormatConfig => {
                 timeZone: state.timeZone || TIMEZONE_CLIENT,
                 forceTZ: !!state.timeZone,
                 dateTime: state.dateTimeFormat || "yyyy-MM-dd HH:mm:ss zzz",
-                date: state.dateFormat || "yyyy/MM/dd",
+                date: state.dateFormat || "yyyy-MM-dd",
                 number: state.numberFormat || "%f",
             } as FormatConfig),
         [state.timeZone, state.dateFormat, state.dateTimeFormat, state.numberFormat]
--- a/gui/src/utils/hooks.ts
+++ b/gui/src/utils/hooks.ts
@@ -105,9 +105,10 @@ export const useFormatConfig = (): FormatConfig => {
                 timeZone: state.timeZone || TIMEZONE_CLIENT,
                 forceTZ: !!state.timeZone,
                 dateTime: state.dateTimeFormat || "yyyy-MM-dd HH:mm:ss zzz",
+                date: state.dateFormat || "yyyy/MM/dd",
                 number: state.numberFormat || "%f",
             } as FormatConfig),
-        [state.timeZone, state.dateTimeFormat, state.numberFormat]
+        [state.timeZone, state.dateFormat, state.dateTimeFormat, state.numberFormat]
     );
 };
 
--- a/src/taipy/gui/utils/date.py
+++ b/src/taipy/gui/utils/date.py
@@ -19,7 +19,7 @@ from pytz import utc
 from .._warnings import _warn
 
 
-def _date_to_ISO(date_val: t.Union[datetime, date, time]) -> str:
+def _date_to_string(date_val: t.Union[datetime, date, time]) -> str:
     if isinstance(date_val, datetime):
         # return date.isoformat() + 'Z', if possible
         try:
@@ -31,7 +31,7 @@ def _date_to_ISO(date_val: t.Union[datetime, date, time]) -> str:
     return date_val.isoformat()
 
 
-def _ISO_to_date(date_str: str) -> t.Union[datetime, date]:
+def _string_to_date(date_str: str) -> t.Union[datetime, date]:
     # return datetime.strptime(date_str, '%Y-%m-%dT%H:%M:%S.%fZ')
     # return datetime.fromisoformat(date_str)
     date = parser.parse(date_str)
--- a/src/taipy/gui/utils/date.py
+++ b/src/taipy/gui/utils/date.py
@@ -9,6 +9,7 @@
 # an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the License for the
 # specific language governing permissions and limitations under the License.
 
+import re
 import typing as t
 from datetime import date, datetime, time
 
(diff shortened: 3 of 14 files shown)


gui/src/components/Taipy/DateSelector.spec.tsx
```diff
@@ -143,7 +143,7 @@ describe("DateSelector Component", () => {
             await userEvent.type(input, "{ArrowLeft}{ArrowLeft}{ArrowLeft}01012001", { delay: 1 });
             expect(dispatch).toHaveBeenLastCalledWith({
                 name: "",
-                payload: { value: "2001-01-01T00:00:00.000Z" },
+                payload: { value: "Mon Jan 01 2001" },
```
FredLL-Avaiga, line 146: should we not format date as YYYMMDD ?

src/taipy/gui/utils/date.py
```diff
@@ -30,7 +31,9 @@ def _date_to_ISO(date_val: t.Union[datetime, date, time]) -> str:
     return date_val.isoformat()
 
 
-def _ISO_to_date(date_str: str) -> datetime:
+def _ISO_to_date(date_str: str) -> t.Union[datetime, date]:
     # return datetime.strptime(date_str, '%Y-%m-%dT%H:%M:%S.%fZ')
     # return datetime.fromisoformat(date_str)
-    return parser.parse(date_str)
+    date = parser.parse(date_str)
+    date_regex = r"^[A-Z][a-z]{2} (?:Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) \d{2} \d{4}$"
```
FredLL-Avaiga, line 38: what happens if the browser local has an impact on toDateString ?

FredLL-Avaiga (approved):
I think we should use a local agnostic date format like YYYMMDD

src/taipy/gui/utils/date.py
dinhlongviolin1: it should not according to mdn as it is compatible with all browsers

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toDateString

gui/src/components/Taipy/DateSelector.spec.tsx
FabienLelaquais: Indeed we should be using ISO 8601 (=> "YYYYMMDD") whenever a date is serialized

dinhlongviolin1: this is the default format of toDateString(). if you think it is better to change it to another format then I will do it :D

gui/src/utils/hooks.ts
```diff
@@ -105,9 +105,10 @@ export const useFormatConfig = (): FormatConfig => {
                 timeZone: state.timeZone || TIMEZONE_CLIENT,
                 forceTZ: !!state.timeZone,
                 dateTime: state.dateTimeFormat || "yyyy-MM-dd HH:mm:ss zzz",
+                date: state.dateFormat || "yyyy/MM/dd",
```
FabienLelaquais: I suspect the proper format should be "yyyy-MM-dd" (homogeneous with the datetime format)

gui/src/components/Taipy/DateSelector.spec.tsx
FredLL-Avaiga: indeed toDateString is always in english ...
If python knows how to parse this by default, I think that we're good

src/taipy/gui/utils/date.py
FabienLelaquais: If input is ISO (as the function name infers) then this pattern should never occur

dinhlongviolin1: Oh, do you mean that the browser timezone would affect the date parsing

tests/taipy/gui/e2e/test_timezone.py
```diff
@@ -65,3 +65,18 @@ def _timezone_test_template(page: "Page", gui: Gui, helpers, time_zone, texts):
     page.wait_for_selector("#text1")
     text1 = page.query_selector("#text1")
     assert text1.inner_text() in texts
+
+
+def test_date_only(page: "Page", gui: Gui, helpers):
+    page_md = """
+<|{t}|id=text1|>
+"""
+    t = _ISO_to_date("Wed Jul 28 1993")  # noqa: F841
```
FabienLelaquais: That is not ISO... or the function is wrongly named, and ENG-locale dependent... which is bad

gui/src/components/Taipy/DateSelector.spec.tsx
dinhlongviolin1: It knows pretty well. Tested it and it works like a charm

tests/taipy/gui/e2e/test_timezone.py
```diff
@@ -65,3 +65,18 @@ def _timezone_test_template(page: "Page", gui: Gui, helpers, time_zone, texts):
...
+    gui._set_frame(inspect.currentframe())
+    gui.add_page(name="test", page=page_md)
+    helpers.run_e2e(gui)
+    page.goto("./test")
+    page.expect_websocket()
+    page.wait_for_selector("#text1")
+    text1 = page.query_selector("#text1")
+    assert text1.inner_text() in ["1993/07/28"]
```
FabienLelaquais: 1993-07-28 (?)

FabienLelaquais (changes requested):
Make sure ISO is used all over the place - we can discuss that anytime.

src/taipy/gui/utils/date.py
FredLL-Avaiga: it could
and if you use YYY-MM-DD you can simply test the length

FabienLelaquais (approved):
👌

src/taipy/gui/utils/date.py
dinhlongviolin1: right now, the toDateString method is the one being used when withTime = False so we dont have to concern with other cases such as YYYY-mm-dd. For the affect of tz on the date parsing, I trust that the date selector would do the job of getting the correct date :D

Issue #939, opened 2023-09-29
Allow GUI action functions that take a callback to use the function itself directly


FabienLelaquais:
**Description**

A typical example of that limitation is the `download()` GUI action: its *on_action* parameter expects the name of the callback function where it would be far simpler to use the function value itself:

`download(state, content=..., on_action="my_callback_name")` is not really user-friendly compared to `download(state, content=..., on_action=my_callback_name)`

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #940, opened 2023-09-29
on_action can be a function (#939)

FredLL-Avaiga:
Addresses #939.

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -66,6 +66,7 @@ from .utils import (
     _get_broadcast_var_name,
     _get_client_var_name,
     _get_css_var_value,
+    _get_expr_var_name,
     _get_module_name_from_frame,
     _get_non_existent_file_path,
     _get_page_from_module,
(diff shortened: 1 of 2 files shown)


FabienLelaquais (approved):
Eaaaasy! 🙏

src/taipy/gui/gui.py
```diff
@@ -1600,7 +1601,11 @@ def _is_broadcasting(self) -> bool:
         except RuntimeError:
             return False
 
-    def _download(self, content: t.Any, name: t.Optional[str] = "", on_action: t.Optional[str] = ""):
+    def _download(self, content: t.Any, name: t.Optional[str] = "", on_action: t.Optional[t.Union[str, t.Callable]] = ""):
+        if callable(on_action):
+            on_action_name = _get_expr_var_name(on_action.__name__)
```
dinhlongviolin1: on_action.__name__ might be undefined

dinhlongviolin1 (approved):
First review on new phone!

src/taipy/gui/gui.py
FabienLelaquais: Didn't think of this one. Lambdas?

FabienLelaquais:
> First review on new phone!

😜

src/taipy/gui/gui.py
FredLL-Avaiga: When would that be ?

PR #942, opened 2023-10-01
update flask version to 3.0

dinhlongviolin1:


src/taipy/gui/server.py
```diff
@@ -95,7 +95,7 @@ def __init__(
 
         # set json encoder (for Taipy specific types)
         self._flask.json_provider_class = _TaipyJsonProvider
-        self._flask.json = self._flask.json_provider_class(self._flask)
+        self._flask.json = self._flask.json_provider_class(self._flask)  # type: ignore
```
FredLL-Avaiga, line 98: good fix 😂

FredLL-Avaiga (approved):

src/taipy/gui/server.py
dinhlongviolin1: It was a dumb error in the first place :D it requires "App". But Flask instance inherits "App" so...

dinhlongviolin1: Nothing we can do unless we fix mypy 😀

FabienLelaquais (approved):
👌

PR #944, opened 2023-10-03
fix default width for layout children

FredLL-Avaiga:
css magic from @JaufreyTpy


ghost (approved):
👌 You are a wizard, Harry.

FabienLelaquais (approved):
🤔

Issue #945, opened 2023-10-04
BUG-Gui Builder error on reload


dinhlongviolin1:
**Description**
When a taipy gui builder page is reloaded, the page is not accessible due to some parsing issue.

**How to reproduce**
A code fragment
```
import pandas as pd
from datetime import datetime
import src.taipy.gui.builder as tgb
from src.taipy.gui import Gui

start_date = datetime(2023, 1, 1)
end_date = datetime(2023, 2, 1)
date_range = pd.date_range(start=start_date, end=end_date)

df = pd.DataFrame({'Date': date_range})
df["Index"] = df.index

text = 10

def on_edit(state, var_name, action, payload):
    print(var_name, action, payload)

with tgb.Page() as page:
    tgb.text("{text}")
    tgb.table(data="{df}", editable=True, filter=True, on_edit=on_edit)

Gui(page=page).run()
```

**Expected behavior**
User should be able to reload the page

**Screenshots**
When available and relevant, screenshots better help show the problem.

**Runtime environment**
Please specify relevant indications.
 - OS: [e.g. Linux, Windows] and version
 - Browser: [e.g. Chrome, Edge, Safari] and version
and any other relevant information.

PR #946, opened 2023-10-04
Fix Gui Builder reload issue (#945)

dinhlongviolin1:
Solve bug #945 

```
import pandas as pd
from datetime import datetime
import src.taipy.gui.builder as tgb
from src.taipy.gui import Gui

start_date = datetime(2023, 1, 1)
end_date = datetime(2023, 2, 1)
date_range = pd.date_range(start=start_date, end=end_date)

df = pd.DataFrame({'Date': date_range})
df["Index"] = df.index

text = 10

def on_edit(state, var_name, action, payload):
    print(var_name, action, payload)

with tgb.Page() as page:
    tgb.text("{text}")
    tgb.table(data="{df}", editable=True, filter=True, on_edit=on_edit)

# page = """
# <|{df}|table|on_edit=on_edit|editable=True|filter=True|>
# """

Gui(page=page).run()
```

### DIFF
--- a/src/taipy/gui/_renderers/_class_api/element_api.py
+++ b/src/taipy/gui/_renderers/_class_api/element_api.py
@@ -11,6 +11,7 @@
 
 from __future__ import annotations
 
+import copy
 import typing as t
 from abc import ABC, abstractmethod
 from collections.abc import Iterable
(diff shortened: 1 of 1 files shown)


FredLL-Avaiga (approved):

FabienLelaquais (approved):
🤔no bad outcome to expect, in terms of perfs?
I guess not too much considering the objects at play.

dinhlongviolin1:
> 🤔no bad outcome to expect, in terms of perfs? I guess not too much considering the objects at play.

I don't think it will affect perfs at all. It will rarely be executed anyway.

Issue #941, opened 2023-10-01
Simplify and clarify 'on_action'


FabienLelaquais:
**Description**

As of now, the generic 'on_action' callback is documented (in the `Gui` class) as expecting three parameters: the state, the invoking element identifier, and a payload dictionary providing more context for the callback.
This to me, looks fine.

The actual implementation is different, though: an additional parameter 'action' is inserted before the `payload` parameter, storing the callback name.
This seems to have been introduced when implementing `Gui.__get_on_cancel_block_ui()`  where instead of using the payload's 'action' property to store the callback name, is was passed directly, then a hack in `Gui.__on_action()` would retrieve it.

I think this is wrong:
- the user rarely would be interested in the callback name value: he already knows where the code is implemented.
- the current implementation is wrong with respect to the documentation

I suggest we should change the code to reflect what the doc says: the action name would be a property of the payload so it can be retrieved if really needed (I don't see a realistic use case at this point). If the callback is a lambda function, the 'action' key of the payload should not be set.

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #947, opened 2023-10-04
Simplify and unify on_action signature

FabienLelaquais:
Addresses #941

+ Doc improvements and fixes


src/taipy/gui/gui.py
```diff
@@ -1238,7 +1237,7 @@ def __get_on_cancel_block_ui(self, callback: t.Optional[str]):
         def _taipy_on_cancel_block_ui(guiApp, id: t.Optional[str], payload: t.Any):
             if _hasscopeattr(self, Gui.__UI_BLOCK_NAME):
                 _setscopeattr(self, Gui.__UI_BLOCK_NAME, False)
-            self.__on_action(id, callback)
+            self.__on_action(id, {"action": callback})
```
FredLL-Avaiga: interesting: should self be guiApp here ?

src/taipy/gui/gui.py
```diff
@@ -1636,6 +1635,7 @@ def _hold_actions(
         message: t.Optional[str] = "Work in Progress...",
     ):  # pragma: no cover
         action_name = callback.__name__ if callable(callback) else callback
+        # TODO: what if lambda? (it does work)
```
FredLL-Avaiga, line 1638: it does work as long as there is only one declaration (the other lambda are named by their code :-)

FredLL-Avaiga (approved):
LGTM
Interesting point: no impact on the FE ...

src/taipy/gui/gui.py
FabienLelaquais: Hmmmm I guess you're right!
Lucky we are that no one is creating multiple Gui instances hu?

src/taipy/gui/gui.py
FabienLelaquais: True.

PR #960, opened 2023-10-09
navbar default property

FredLL-Avaiga:


src/taipy/gui/viselements.json
```diff
@@ -758,8 +758,9 @@
           "properties": [
             {
                 "name": "lov",
+                "default_property": true,
                 "type": "dict[str, any]",
```
FabienLelaquais, line 762: Why isn't this `dict[str, str]`, reading the doc?

FabienLelaquais (approved):
👌

src/taipy/gui/viselements.json
FredLL-Avaiga: its str or icon + str ...

FabienLelaquais: Ok so the doc is, say incomplete...

FredLL-Avaiga: same as lovComp.lov

Issue #968, opened 2023-10-10
BUG-Cant use state within notebook


dinhlongviolin1:
**Description**
State cannot be accessed within the context of notebook. This is due to the recent changes in page scope management within the State class.

PR #969, opened 2023-10-10
Fix state access within notebook (#968)

dinhlongviolin1:
Solve issue #968 for Avaiga/taipy-doc#521


FabienLelaquais (approved):
Oh Lord. 😶

FredLL-Avaiga (approved):
I guess it's ok

PR #970, opened 2023-10-12
allow optional lov_value property (for taipy)

FredLL-Avaiga:
This will allow to not show a warning when an optional lov_value is not bound (think Datanode viewer => scenario)


dinhlongviolin1 (approved):

FabienLelaquais (approved):
Makes sense!

Issue #971, opened 2023-10-12
BUG - Table Visual Element display boolean values as clickable


AlexandreSajus:
The table visual element displays boolean values using an on-off slider:
![image](https://github.com/Avaiga/taipy-gui/assets/62465003/ce091425-752a-4175-9098-fc7740b2372e)

This is a good idea and it looks good. Unfortunately, it is clickable, and clicking it does not do anything. We should remove the clickable property

PR #974, opened 2023-10-13
Cursor on table (non edit mode)

FredLL-Avaiga:
Addresses #971

So that the UI feedback doesn't lure the user to think it's editable


dinhlongviolin1 (approved):
LGTM

FabienLelaquais (approved):
🤔

PR #981, opened 2023-10-19
reset module context after invoking callback + use context manager (#990)

dinhlongviolin1:
Solve Issue #990

### DIFF
--- a/src/taipy/gui/gui.py
+++ b/src/taipy/gui/gui.py
@@ -510,7 +510,7 @@ class Gui:
                 res = self._bindings()._get_or_create_scope(message.get("payload", ""))
                 client_id = res[0] if res[1] else None
             self.__set_client_id_in_context(client_id or message.get(Gui.__ARG_CLIENT_ID))
-            with self._ctx_locals_context(message.get("module_context") or None):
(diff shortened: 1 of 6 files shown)


src/taipy/gui/gui.py
```diff
@@ -1138,7 +1138,10 @@ def _call_user_callback(
                 self.__set_client_id_in_context(state_id)
                 if module_context is not None:
                     self._set_locals_context(module_context)
-                return self._call_function_with_state(user_callback, args)
+                callback_result = self._call_function_with_state(user_callback, args)
```
FabienLelaquais: Wasn't a 'finally' statement ever more clear?

FabienLelaquais (approved):
Fine with me

src/taipy/gui/gui.py
FredLL-Avaiga: and it would allow for exception management

FredLL-Avaiga (changes requested):
definitely "finally"

FabienLelaquais (approved):
Indeed faaar more readable.
Well done!

FabienLelaquais (approved):
Nice!

Issue #976, opened 2023-10-16
BUG- Weird interaction between clients with invoke long callback


FlorianJacta:
**Description**

When two applications run an *invoke_long_callback*; there are strange behaviors with the client. 

```console
TaipyGuiWarning: Empty session id, using global scope instead.
TaipyGuiWarning: Empty session id, using global scope instead.
TaipyGuiWarning: Empty session id, using global scope instead.
TaipyGuiWarning: Empty session id, using global scope instead.
```

**How to reproduce**

Run this code and observe the value of each client. The number should always rise in both.

```python
from taipy.gui import Gui, invoke_long_callback, notify, navigate
import time 

page = """
<|{value}|>
"""

def update(state, status):
    state.value += 1

def iddle():
     while True:
        time.sleep(1)

def on_init(state):
    invoke_long_callback(state, iddle, [], update, [], 1000)

value = 0

if __name__ == "__main__":
    gui = Gui(page=page)
    gui.run(title='Raising number')

```

**Expected behavior**
Each context should be kept from the others.

**Screenshots**
![pb_client](https://github.com/Avaiga/taipy-gui/assets/98709993/03443035-59cd-49e1-a924-a89dd165e71a)

**Runtime environment**
Taipy GUI 3.0.dev4

dinhlongviolin1:
I have tested this one. I could not replicate the issue and it might be on your side due to the missing session id. Please reopen the issue if the error still persisted on your side. Here is the code that I used.

```
from src.taipy.gui import Gui, invoke_long_callback
import time

page = """
<|{value}|>
"""


def update(state, status):
    state.value += 1


def iddle():
    while True:
        time.sleep(0.5)


def on_init(state):
    invoke_long_callback(state, iddle, [], update, [], 500)


value = 0

if __name__ == "__main__":
    gui = Gui(page=page)
    gui.run(title="Raising number", server_config={"socketio": {"ping_interval": 0.5}})
```

FlorianJacta:
Not working with this code. I don't think this is on my side. It is replicated even on Taipy Cloud. Have you created two different clients and have yet to see the weird changes?

dinhlongviolin1:
> Not working with this code. I don't think this is on my side. It is replicated even on Taipy Cloud. Have you created two different clients and have yet to see the weird changes?

I have tried on more than 3 separate clients on 3 browsers (chrome, safari, firefox) at the same time and it works perfectly on my side.

[...]

PR #983, opened 2023-10-19
Initialize client id if need before anything else (#976)

FredLL-Avaiga:
Fix issue #976


dinhlongviolin1 (approved):
🥇 🥇 🥇

src/taipy/gui/gui.py
```diff
@@ -460,9 +460,12 @@ def _get_client_id(self) -> str:
             else getattr(g, Gui.__ARG_CLIENT_ID, "unknown id")
         )
 
-    def __set_client_id_in_context(self, client_id: t.Optional[str] = None):
+    def __set_client_id_in_context(self, client_id: t.Optional[str] = None, force=False):
         if not client_id and request:
             client_id = request.args.get(Gui.__ARG_CLIENT_ID, "")
+        if not client_id and force:
+            res = self._bindings()._get_or_create_scope("")
```
dinhlongviolin1, line 467: Instead of an empty string, should this be a defined value

src/taipy/gui/server.py
```diff
@@ -102,8 +102,9 @@ def __init__(
         self._is_running = False
 
         # Websocket (handle json message)
+        # adding args for the one call with a server ack request
         @self._ws.on("message")
-        def handle_message(message) -> None:
+        def handle_message(message, *args) -> None:
```
dinhlongviolin1, line 107: I dont see that the args is being passed into any other function. Am I missing something here

src/taipy/gui/gui.py
```diff
@@ -954,7 +959,8 @@ def _send_ws_id(self, id: str) -> None:
             {
                 "type": _WsType.CLIENT_ID.value,
                 "id": id,
-            }
+            },
+            allow_grouping=False,
```
dinhlongviolin1, line 963: Nice one!

dinhlongviolin1 (approved):

src/taipy/gui/server.py
FredLL-Avaiga: I use the Ack server callback and it complained once that there were 2 arguments

src/taipy/gui/gui.py
FredLL-Avaiga: nope
force is set only when called from _init_route
- if there's a client id, we won't go through this
- else we want a new id

Which should never happen as we're waiting a return before showing the page for the first time

dinhlongviolin1: thanks. Make sense to me

FabienLelaquais (approved):
Excellent pick!

PR #988, opened 2023-10-20
Doc for html() builder class

FabienLelaquais:
+ Support for text nodes
+ Simplify tests (Gui without run())

### DIFF
--- a/src/taipy/gui/builder/_element.py
+++ b/src/taipy/gui/builder/_element.py
@@ -155,8 +155,7 @@ class html(_Block):
         if self._ELEMENT_NAME:
             attrs = ""
             if self._properties:
-                attrs = " "+" ".join([f'{k}="{str(v)}"' for k, v in self._properties.items()]) 
-            open_tag = f""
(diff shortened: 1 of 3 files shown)


src/taipy/gui/builder/_element.py
```diff
@@ -106,24 +106,46 @@ def __init__(self, *args, **kwargs):
...
+
+        Examples:
+            - To generate `<br/>`, use:
+               ```
+               html("br")
+               ```
+            - To generate `<h1>My page title</h1>`, use:
+               ```
+               html("h1", "My page title")
+               ```
```
FredLL-Avaiga, line 135: No example of using kwargs?

FredLL-Avaiga (approved):

src/taipy/gui/builder/_element.py
FabienLelaquais: Well I had an example with 'id', but for some reason it's gone.
Thanks!

FabienLelaquais: Now I know why. Friday evening side effect.
Thanks again

dinhlongviolin1 (approved):
LGTM

PR #991, opened 2023-10-22
Minor fixes

dinhlongviolin1:
- Check port before starting proxy in notebook (bug discovered while in call with Fabien)
- Use try-finally for broadcasting

### DIFF
--- a/src/taipy/gui/utils/proxy.py
+++ b/src/taipy/gui/utils/proxy.py
@@ -27,6 +27,7 @@ from twisted.web.proxy import ProxyClient, ProxyClientFactory
 from twisted.web.resource import Resource
 from twisted.web.server import NOT_DONE_YET, Site
 
+from .is_port_open import _is_port_open
 from .singleton import _Singleton
 
 if t.TYPE_CHECKING:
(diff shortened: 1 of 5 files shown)


src/taipy/gui/utils/proxy.py
```diff
@@ -93,10 +94,16 @@ def __init__(self, gui: "Gui", listening_port: int) -> None:
...
         if self._is_running:
             return
-        self._is_running = True
-        site = Site(_TaipyReverseProxyResource(self._gui._get_config("host", "127.0.0.1"), b"", self._gui))
-        reactor.listenTCP(self._listening_port, site)
+        host = self._gui._get_config("host", "127.0.0.1")
+        port = self._listening_port
+        if _is_port_open(host, port):
+            raise ConnectionError(
+                f"Port {port} is already opened on {host}. You have another server application running on the same port."
```
FabienLelaquais, line 101: So that is going to be handy.

FabienLelaquais (approved):
👌

src/taipy/gui/utils/proxy.py
dinhlongviolin1: Indeed. You found the bug while playing around with it 😀

Issue #503, opened 2022-11-28
Chart: support candlestick traces


FredLL-Avaiga:
**Description**
Support for Plotly candlestick trace 
https://plotly.com/javascript/candlestick-charts/

**Acceptance Criteria**
- [ ] Ensure new code is unit tested, and check code coverage is at least 90%
- [ ] Propagate any change on the demos and run all of them to ensure there is no breaking change
- [ ] Ensure any change is well documented

PR #1031, opened 2023-11-30
Table row selection (#503)

FredLL-Avaiga:
col is None if the user clicked outside the column text

```
def on_table_action(state: State, var_name: str, payload: dict):
    print(f"on_table_action(var_name={var_name}, payload={payload})")

```

```
on_table_action(var_name=perf_df, payload={'action': 'on_table_action', 'index': 0, 'col': 'Scenario', 'args': []})
on_table_action(var_name=perf_df, payload={'action': 'on_table_action', 'index': 0, 'col': None, 'args': []})
```


dinhlongviolin1 (approved):
Lgtm

FabienLelaquais (approved):
LGTM

PR #1033, opened 2023-12-01
Adding comments to Taipy pages definition

PolicarpoBatistaUliana:
The included _comment.py python file contain two rotins that allow the use of comentes in Taipy pages strings  that using Markdown codes. 
For HTML codes the comments are made using / *    * / delimiters, for Markdown
we define the same but including the use of a like python comment given by "/#" delimiter
Two types of comments are allowed inside of the string that define the page Markdown code::
A- Line-ending comment: /# XXXX  =>  Need be used in the end of the line or in a complete line. All words before /# will be removed  
B- Comment inserted: / * XXXX  * /    =>  Can be used inside the code. 	                                                                    Only the text  / * XXXX  * /   will be removed

Example of the new use comments in Taipe pages definition:

page_login ="""
/ * The Taipy page beginning  here,  to get 3 new lines use \n to do this:* /  \n\n\n
/# In the line below the msg1[1] contain "Username or Email"
/#presented in 8 idioms:   
/#
/# "Username or Email" message,  in 8 idioms:
<|{msg1[Idioma_sel.id / *number of selected idiom =1 to 8 * /]}|text|raw|>\n
/#
/#This Markdown get  the user name
<|{UserName}|input|> \n 
/#
/# "Inform Password" message, in 8 idions:
<|{msg2[Idioma_sel.id]/* Password */}|text|raw|>:\n
/#
/#This Markdown get  the Password:
<|{PassWord}|input|> \n 
/# We need a way to hide the letters
/# when the user enters this value!
/#
/#This Markdown creat a Login button: 
<|msg3[Idioma_sel.id+1]/ *button caption: "Login" in 8 idioMs * /}|button|active|
on_action=confirma_login / * python rotine that react to the button press action* /|>\n
/# We need define a rotine in python to use this UserName and PassWord values:
/# def  confirma_login(state):
/#      print("nome User=",state.UserName ); 
/#      print("senha=",state.PassWord ); 
/#
/ * The Taipy page finish here, with this  3'', the use of new line is optional * / \n 
"""
OBS: I use /"space"*  because /* and */  (/"no-space"*  and   *"no-space"/ ) is given problem in it chat...
 
The same page with  Markdown code, after the comments be removed:

page_login =""" 
\n\n\n
<|{msg1[Idioma_sel.id]}|text|raw|>\n
<|{UserName}|input|> \n 
<|{msg2[Idioma_sel.id]}|text|raw|>:\n
<|{PassWord}|input|> \n 
<|msg3[Idioma_sel.id+1]}|
button|active|on_action=confirma_login|>\n
"""
Observation: Whiteout comment is a more clean code, but hard to undusted 
this msg1 to msg3 meaning, and the Idioma_sel use,  and that 
confirma_login is a python function name....
Another benefit of this scheme is the possibility of disabling parts of the code for testing purposes.

I believe this is a very positive feature and nothing is lost with it, with the fact that you can post comments and can easy disabling code parts whitout lost of information. 
If the user don't want comments, he  just don't use it...
Now, if this improvement is accepted, it is necessary to modify the documentation to show this new possibility of making comments in Taipy.
I can help do this if necessary....
Best Regards
Dr. Policarpo Yoshin Ulianov.
poliyu77@gmail.com
PS: I did basic tests with about 15 types of pages on Linux Debian 11 and Windows 10, with Python 3.11 and this scheme worked well. I believe it shouldn't cause any problems as it's a very simple routine that just removes comments, but it would be interesting for you to test it a little better.

### DIFF
--- a/src/taipy/gui/_comment.py
+++ b/src/taipy/gui/_comment.py
@@ -1,104 +1,64 @@
 """
-This initial text  must be removed and placed in the Taipy documentetion:
+This Python file contains two routines that enable the use of comments in Taipy pages using Markdown codes. It supports two types of comments:
+ A- Line-ending comment (/#): Should be used at the end of a line or on a standalone line. Everything after /# will be removed.
+ B- Inserted comment (/* ... */): Can be used within the code. Only the text within /* ... */ will be removed.
 
-This Python file contain two rotines that allow the use of comentes in Taipy peges
-that using Markdown codes. 
-For HTML codes the comments are made using /*    */ delimiters, for Markdown
-we define the same but incliding the use of a like python comment given by /# delimiter
+Example usage in Taipy pages definition:
 
-Two types of comments are allowed inside of the string that define the page Markdown code::
-	 A- Line-ending comment: /# XXXX  =>  Need be used in the end of the line or in a complete line.
-	                                                                    All words before /# will be removed  
-	 B- Comment inserted: /* XXXX  */    =>  Can be used inside the code. 
-	                                                                    Only the text  /* XXXX  */   will be removed
-
-Example of the new use comments in Taipe pages definition:
-
-page_login ='''''' 
-/* The Taipy page begining  here,  to get 3 new lines use \n to do this:*/  \n\n\n
-/# In the line below the msg1[1] contain "Username or Email"
-/#presented in 8 idions:   
-
-/# "Username or Email" message,  in 8 idions:
-<|{msg1[Idioma_sel.id /*number of selected idiom =1 to 8 */]}|text|raw|>\n
-
-/#This Markdown get  the user name
+Original Page with Comments:
+---------------------------
+page_login = '''
+/* Page starts here and two new lines will be include:*/ \n\n
+/# This line will be removed and this 3 new lines not has efect: \n\n\n 
+/# ''Username or Email'' message in various languages:
+<|{msg1[Idioma_sel.id/* idiom number from 0 to 7*/]}|text|raw|>
+\n /# This new line will be usede because '\n' is place berore of '/#'
+/* These lines of code was disabled for  purpose of testing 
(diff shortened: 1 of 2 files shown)


FlorianJacta:
Thank you for this PR @PolicarpoBatistaUliana ! It seems like a nice addition to Taipy. We are currently merging taipy, taipy-gui, and taipy-core in one package. We will let you know when this process is done.

FabienLelaquais:
Very interesting proposal indeed, @PolicarpoBatistaUliana!
I really like the idea of commented Markdown, but I want to be very cautious on how this would impact (or not impact) the "regular" parsing of Markdown: besides integrating Taipy GUI elements, we try not to impact the MD syntax at all, to allow for reading regular MD content.
One issue I can see, for example, is that a user creates a piece of pre-formatted code (say C code), then all the C comments would be removed.

Another issue (easier to deal with) is the fact that in Taipy GUI, pages can be added one by one, and not only when initializing a Gui instance. And pages can have different types, not only Markdown (see HTML or the Page Builder API).

I like the feature though, and your implementation is neat.

Now I propose that I create an issue on this topic: we want to decide on a syntax that would impact Markdown parsing less, and process the Taipy GUI comments at a code location where we know we're dealing with Markdown, and nothing else.
What do you think?

FabienLelaquais:
And yes, as @FlorianJacta mentioned, we are just in the middle of merging all the Taipy packages into the 'taipy' repository, where all the code, issues, PRs will be or are transferred. By the way, that's one of the reasons why you've had to wait for our answer.

I propose that, if you agree, I create an issue (or you can) in the `taipy` repository, repeating what you propose here including the code, then work on a safer syntax, yet providing this valuable functionality.

FabienLelaquais:
Captured in https://github.com/Avaiga/taipy/issues/835

PR #1035, opened 2023-12-14
Change value name for dialog

### LINKED ISSUE #382: BUG-Using "open" as the value for a dialog visual element as the documentation suggests fails
**Description**
Using `open` as the main value for a dialog visual element will make the dialog permanently opened. This might be because open is a Python function available by default.

**How to reproduce**
```python
from taipy.gui import Gui, State, notify

open = True

modal_dialog = """
<|{open}|dialog|title=Are you a scientist?|width=30%|
**Warning**
You must be a scientist to be able to view this page
<br/>
<|I am a scientist|button|class_name=fullwidth plain|on_action=close_modal_window_and_render_page|>
|>

<|part|render={not open}|
All the sciency stuff now shows up here
|>
"""


def close_modal_window_and_render_page(state: State):
    state.open = False
    notify(state, "S", "Welcome, Scientist!")


gui = Gui(page=modal_dialog)

gui.run()
```

This code generates a dialog box that can't be closed (no errors in logs)
image

Replacing `open` with `opening` fixes the issue.

**Expected behavior**
Either change the name of `open` in the documentation as an easy fix or find a way around this issue in the code

**Runtime environment**
 - OS: Windows 11
 - Browser: Chrome
 - Taipy: 3.0.0

AlexandreSajus:
Fixes https://github.com/Avaiga/taipy/issues/382

This is my first PR in GUI, I inferred that this would be the correct change, but please make sure this is functional

### DIFF
--- a/src/taipy/gui/viselements.json
+++ b/src/taipy/gui/viselements.json
@@ -1088,7 +1088,7 @@
         ],
         "properties": [
           {
-            "name": "open",
+            "name": "show_dialog",
             "default_property": true,
             "type": "bool",
             "default_value": "False",


FredLL-Avaiga (changes requested):
- the issue is in the user code
- viselements.json is used for doc and has no impact on the code
- taipy-gui repo is/will soon be archived (changes should go to taipy repo)

FabienLelaquais (changes requested):
This is not a good idea.
Backward compatibility has to be maintained as long as we can.
Renaming a property has very little added value (except, and I agree, a better understanding of what it does).
However, when naming Taipy GUI properties, we always need to balance the tradeoff between clearer (and longer) names, and code conciseness. 
To me, 'open' is just fine. If 'o' worked, I would have used that.

Bottom line: what's the purpose of that?

